PR #15, opened 2020-04-13
efficientnet feature extractor

qiminchen:
New FeatureExtractor
- [x] Implement a FeatureExtractor in `spacer/extract_features.py`

Utils
- [x] `torch_utils.py`
- [x] `extract_features.py`
- [x] `extract_features_utils.py`
- [x] `features_extracting_training.py`
- [x] `scripts_utils.py`
- [x] `efficientnet.py`
- [x] `efficientnet_utils.py`

Unit test
- [x] `test_torch_utils.py`
- [x] `test_extract_features.py`
- [x] `test_extract_features_utils.py`
- [x] `test_models.py`

Typing hints and PEP8 cleanup

- scripts
 
  - [x] `classifier_training_regression.py`
  - [x] `features_extracting_training.py`
  - [x] `scripts_utils.py`

- spacer

  - [x] `caffe_utils.py`
  - [x] `torch_utils.py`
  - [x] `extract_features_utils.py`
  - [x] `train_utils.py`

### DIFF
--- a/scripts/features_extracting_training.py
+++ b/scripts/features_extracting_training.py
@@ -29,11 +29,15 @@ class ExtractFeatures:
     2. Load images and anns.meta from s3
     3. Extract features and save to local
     """
-    def __init__(self, extractor_name, local_path):
+    def __init__(self,
+                 extractor_name: str,
+                 local_path: str) -> None:
         self.extractor_name = extractor_name
         self.local_path = local_path
 
-    def __call__(self, source_id, export_name='beta_export'):
+    def __call__(self,
+                 source_id: int,
+                 export_name: str = 'beta_export') -> None:
         conn = config.get_s3_conn()
         bucket = conn.get_bucket('spacer-trainingdata', validate=True)
 
@@ -84,11 +88,15 @@ class TrainClassifier:
     1. Download meta.json and anns.json to local
     2. Train the classifier
     """
-    def __init__(self, local_path):
+    def __init__(self,
+                 local_path: str) -> None:
         self.local_path = local_path
 
     @staticmethod
-    def _cache(source_id, source_root, feats_root, export_name='beta_export'):
+    def _cache(source_id: int,
+               source_root: str,
+               feats_root: str,
+               export_name: str = 'beta_export') -> None:
         conn = config.get_s3_conn()
         bucket = conn.get_bucket('spacer-trainingdata', validate=True)
 
@@ -121,7 +129,10 @@ class TrainClassifier:
             if not os.path.exists(meta_local):
                 meta_keys[idx].get_contents_to_filename(meta_local)
 
-    def __call__(self, source_id, n_epochs, export_name='beta_export'):
+    def __call__(self,
+                 source_id: int,
+                 n_epochs: int = 5,
+                 export_name: str = 'beta_export') -> None:
 
         print('-> Downloading data for source id: {}.'.format(source_id))
         source_root = os.path.join(self.local_path, 's{}'.format(source_id))
@@ -140,7 +151,10 @@ class TrainClassifier:
         start_training(source_root, train_labels, val_labels, n_epochs)
 
 
-def main(extractor_name, source_id, n_epochs, local_path):
+def main(extractor_name: str,
+         source_id: int,
+         n_epochs: int,
+         local_path: str) -> None:
     extractor = ExtractFeatures(extractor_name, local_path)
     extractor(source_id)
     classifier = TrainClassifier(local_path)
--- a/scripts/features_extracting_training.py
+++ b/scripts/features_extracting_training.py
@@ -4,19 +4,23 @@ This file is used for:
     2. train the classifier
 on the exported spacer training data and confirms that
 it get similar (or higher) performance than production.
+
+To use do:
+python scripts/features_extracting_training.py \
+    efficientnet_b0_ver1 294 10 /path/to/features
 """
 
 import os
+import fire
 import json
-import glob
 from io import BytesIO
 from boto.s3.key import Key
 
 from spacer import config
 from spacer.tasks import extract_features
-from spacer.data_classes import ImageLabels
-from spacer.train_classifier import trainer_factory
 from spacer.messages import DataLocation, ExtractFeaturesMsg
+
+from scripts.scripts_utils import build_traindata, start_training
 
 
 class ExtractFeatures:
@@ -25,46 +29,54 @@ class ExtractFeatures:
     2. Load images and anns.meta from s3
     3. Extract features and save to local
     """
-    def __init__(self, extractor_name):
+    def __init__(self, extractor_name, local_path):
         self.extractor_name = extractor_name
+        self.local_path = local_path
 
     def __call__(self, source_id, export_name='beta_export'):
         conn = config.get_s3_conn()
         bucket = conn.get_bucket('spacer-trainingdata', validate=True)
 
-        print("-> Fetching {} image and annotation meta files...".format(source_id))
-        all_keys = bucket.list(prefix='{}/s{}/images'.format(export_name, source_id))
+        print("-> Fetching {} image and annotation meta files...".format(
+            source_id))
+        all_keys = bucket.list(prefix='{}/s{}/images'.format(
+            export_name, source_id))
         img_keys = [key.name for key in all_keys if key.name.endswith('jpg')]
-        ann_keys = [Key(bucket=bucket, name=img.replace('jpg', 'anns.json')) for img in img_keys]
+        ann_keys = [Key(bucket=bucket, name=img.replace('jpg', 'anns.json'))
+                    for img in img_keys]
 
-        feats_local_path = os.path.join(config.LOCAL_FEATURE_DIR, 's{}'.format(source_id), 'images')
+        feats_local_path = os.path.join(self.local_path, 's{}'.format(
+            source_id), 'images')
         if not os.path.isdir(feats_local_path):
             os.system('mkdir -p ' + feats_local_path)
 
         print("-> Extracting features...")
         for idx, im_key in enumerate(img_keys):
-            anns = json.load(BytesIO(ann_keys[idx].get_contents_as_string()))
-            rowcols = [(ann['row'], ann['col']) for ann in anns]
+            feature_path = os.path.join(
+                feats_local_path,
+                im_key.split('/')[-1].replace('jpg', 'features.json')
+            )
+            if not os.path.exists(feature_path):
+                anns = json.load(BytesIO(
+                    ann_keys[idx].get_contents_as_string()
+                ))
+                rowcols = [(ann['row'], ann['col']) for ann in anns]
 
-            img_loc = DataLocation(storage_type='s3',
-                                   key=im_key,
-                                   bucket_name='spacer-trainingdata')
-            feature_loc = DataLocation(
-                storage_type='filesystem',
-                key=os.path.join(
-                    feats_local_path,
-                    im_key.split('/')[-1].replace('jpg', 'features.json')
+                img_loc = DataLocation(storage_type='s3',
+                                       key=im_key,
+                                       bucket_name='spacer-trainingdata')
+                feature_loc = DataLocation(
+                    storage_type='filesystem',
+                    key=feature_path
                 )
-            )
-
-            msg = ExtractFeaturesMsg(
-                job_token='extract_job',
-                feature_extractor_name=self.extractor_name,
-                rowcols=rowcols,
-                image_loc=img_loc,
-                feature_loc=feature_loc
-            )
-            return_msg = extract_features(msg)
+                msg = ExtractFeaturesMsg(
+                    job_token='extract_job',
+                    feature_extractor_name=self.extractor_name,
+                    rowcols=rowcols,
+                    image_loc=img_loc,
+                    feature_loc=feature_loc
+                )
+                _ = extract_features(msg)
 
 
 class TrainClassifier:
@@ -72,6 +84,9 @@ class TrainClassifier:
     1. Download meta.json and anns.json to local
     2. Train the classifier
     """
+    def __init__(self, local_path):
+        self.local_path = local_path
+
     @staticmethod
     def _cache(source_id, source_root, feats_root, export_name='beta_export'):
         conn = config.get_s3_conn()
@@ -82,10 +97,16 @@ class TrainClassifier:
         if not os.path.isdir(feats_root):
             os.system('mkdir -p ' + feats_root)
 
-        mdkey = bucket.get_key('{}/s{}/meta.json'.format(export_name, source_id))
-        all_keys = bucket.list(prefix='{}/s{}/images'.format(export_name, source_id))
-        anns_keys = [key for key in all_keys if key.name.endswith('.anns.json')]
-        meta_keys = [key for key in all_keys if key.name.endswith('.meta.json')]
+        mdkey = bucket.get_key('{}/s{}/meta.json'.format(
+            export_name, source_id
+        ))
+        all_keys = bucket.list(prefix='{}/s{}/images'.format(
+            export_name, source_id
+        ))
+        anns_keys = [key for key in all_keys
+                     if key.name.endswith('.anns.json')]
+        meta_keys = [key for key in all_keys
+                     if key.name.endswith('.meta.json')]
         assert len(anns_keys) == len(meta_keys)
 
         print("-> Downloading {} metadata...".format(len(anns_keys)))
@@ -103,52 +124,28 @@ class TrainClassifier:
     def __call__(self, source_id, n_epochs, export_name='beta_export'):
 
         print('-> Downloading data for source id: {}.'.format(source_id))
-        source_root = os.path.join(config.LOCAL_FEATURE_DIR, 's{}'.format(source_id))
+        source_root = os.path.join(self.local_path, 's{}'.format(source_id))
         feats_root = os.path.join(source_root, 'images')
-        self._cache(source_id, source_root, feats_root, export_name='beta_export')
+        self._cache(
+            source_id, source_root, feats_root, export_name
+        )
 
         # Create the train and val ImageLabels data structures.
         print('-> Assembling train and val data for source id: {}'.format(
             source_id))
-        files = glob.glob(os.path.join(source_root, 'images', "*.json"))
-        train_labels = ImageLabels(data={})
-        val_labels = ImageLabels(data={})
-        for idx, filename in enumerate(files):
-            if 'anns' in filename:
-                with open(filename, 'r') as fp:
-                    anns = json.load(fp)
-                meta_filename = filename.replace('anns', 'meta')
-                with open(meta_filename, 'r') as fp:
-                    meta = json.load(fp)
-                if meta['in_trainset']:
-                    labels = train_labels
-                else:
-                    assert meta['in_valset']
-                    labels = val_labels
-                labels.data[filename.replace('anns', 'features')] = [
-                    (ann['row'], ann['col'], ann['label']) for ann in anns
-                ]
+        train_labels, val_labels = build_traindata(feats_root)
 
-        feature_loc = DataLocation(storage_type='filesystem', key='')
         # Perform training
         print("-> Training...")
-        trainer = trainer_factory('minibatch')
-        clf, val_results, return_message = trainer(
-            train_labels, val_labels, n_epochs, [], feature_loc)
+        start_training(source_root, train_labels, val_labels, n_epochs)
 
-        with open(os.path.join(source_root, 'meta.json')) as fp:
-            source_meta = json.load(fp)
 
-        print('-> Re-trained {} ({}). Old acc: {:.1f}, new acc: {:.1f}'.format(
-            source_meta['name'],
-            source_meta['pk'],
-            100 * float(source_meta['best_robot_accuracy']),
-            100 * return_message.acc)
-        )
+def main(extractor_name, source_id, n_epochs, local_path):
+    extractor = ExtractFeatures(extractor_name, local_path)
+    extractor(source_id)
+    classifier = TrainClassifier(local_path)
+    classifier(source_id, n_epochs)
 
 
 if __name__ == '__main__':
-    extractor = ExtractFeatures(extractor_name='efficientnet_b0_ver1')
-    extractor(source_id=294)
-    classifier = TrainClassifier()
-    classifier(source_id=294, n_epochs=10)
+    fire.Fire(main)
--- a/scripts/features_extracting_training.py
+++ b/scripts/features_extracting_training.py
@@ -7,14 +7,16 @@ it get similar (or higher) performance than production.
 """
 
 import os
-import tqdm
 import json
+import glob
 from io import BytesIO
 from boto.s3.key import Key
 
 from spacer import config
-from spacer.messages import DataLocation, ExtractFeaturesMsg
 from spacer.tasks import extract_features
+from spacer.data_classes import ImageLabels
+from spacer.train_classifier import trainer_factory
+from spacer.messages import DataLocation, ExtractFeaturesMsg
 
 
 class ExtractFeatures:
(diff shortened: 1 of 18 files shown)


qiminchen:
@beijbom hey Oscar, all the unit tests are passed on my local but The Travis CI build failed b/c of the permission issue, do you know how to solve this?

```======================================================================
FAIL: test_gray (spacer.tests.test_torch_utils.TestClassifyFromPatchList)
----------------------------------------------------------------------
  File "/home/travis/build/beijbom/pyspacer/spacer/tests/test_torch_utils.py", line 42, in setUp
    'efficientnetb0_5eps_best.pt')
  File "/home/travis/build/beijbom/pyspacer/spacer/storage.py", line 186, in download_model
    assert config.HAS_S3_MODEL_ACCESS, "Need access to model bucket."
AssertionError: Need access to model bucket.
======================================================================
FAIL: test_rgb (spacer.tests.test_torch_utils.TestClassifyFromPatchList)
----------------------------------------------------------------------
  File "/home/travis/build/beijbom/pyspacer/spacer/tests/test_torch_utils.py", line 42, in setUp
    'efficientnetb0_5eps_best.pt')
  File "/home/travis/build/beijbom/pyspacer/spacer/storage.py", line 186, in download_model
    assert config.HAS_S3_MODEL_ACCESS, "Need access to model bucket."
AssertionError: Need access to model bucket.
```

spacer/config.py
```diff
@@ -116,6 +116,7 @@ def get_local_model_path():
 
 # Check access to select which tests to run.
 HAS_CAFFE = importlib.util.find_spec("caffe") is not None
+HAS_TORCH = importlib.util.find_spec("torch") is not None
```
beijbom, line 119: We actually don't need this. Torch is available in all builds. It's just caffe that is finicky since only available in Docker.

spacer/torch_utils.py
```diff
@@ -0,0 +1,130 @@
...
+"""
+
+import torch
+import numpy as np
+from spacer import models
+from torchvision import transforms
+from collections import OrderedDict
+
+
+def gray2rgb(im):
```
beijbom, line 12: move this to shared `extract_features` util?

spacer/torch_utils.py
```diff
@@ -0,0 +1,130 @@
...
+    :return: a transformer which takes in a image and return a normalized tensor
+    """
+
+    transformer = transforms.Compose([
+        transforms.ToTensor(),
+    ])
+    return transformer
+
+
+def crop_patch(im, crop_size, scale, point_anns, offsets=None):
```
beijbom, line 33: This also looks similar to the one in `caffe_utils`. So move to shared util.

beijbom:
hey @qiminchen : I had disabled s3 access for Travis due to the SQS tests. But since those are skipped now, i re-enalbled it. But also, you should use the `@unittest.skipUnless(config.HAS_S3_MODEL_ACCESS, 'No access to models')` decorator for tests that require model access, since we can't assume all developers have that access. That will skip the test rather than fail out.

A few more notes looking from glancing at your code:
- Please re-base to master now that I merged my big PR.
- You don't have to follow the exact structure of the caffe feature extractor as long as you implement the interface. For example, you don't have to call your method `classify_from_patchlist`. Although, if it is convenient, by all means use it. But I do think it's pretty confusing for the reader to have two methods that have the same exact name but work on different libraries. Better then to make a new shared `classify_from_patchlist` method that is a shared util agnostic to the feature extractor. Or perhaps make a method `im_to_patchlist` that only does the cropping and pre-processing and then write new methods `classify_from_patchlist` that take in the patchlist. Sorry, this comment got a bit confusing. Perhaps take a stab at finishing this up after rebasing to new master and then I can add more specific comments.
- If there are utility function in `caffe_utils` that you want to use, perhaps move to a shared `extract_features_utils` file? 
- All imports go at the beginning of the file. The only reason I broke this rule is that caffe isn't available in all builds (pip install and clone) so I had to protect those imports.

beijbom:
@qiminchen : CI pipeline permission updated.

beijbom:
One more suggestion @qiminchen : can you write a script (pls. put in in the `scripts` folder), that runs this new feature extractor together with the classifier training on the exported spacer trainingdata and confirms that it get similar (or higher) performance than production. That would function as a regression test basically.

spacer/config.py
qiminchen: Good to know this, Pytorch availability checking is removed from `config.py`

spacer/torch_utils.py
qiminchen: `gray2rgb()` moved to shared `extract_features_utils.py`

spacer/torch_utils.py
qiminchen: `crop_patch()` moved to the shared `extract_features_utils.py`

qiminchen:
> * Please re-base to master now that I merged my big PR.

Done
> * If there are utility functions in caffe_utils that you want to use, perhaps move to a shared extract_features_utils file?

- I moved `gray2rgb()`, `crop_patch()` and `crop_simple()` to the shared `extract_features_utils.py`
- `torch_utils.py` only serves as feature extraction

so in `extract_features.py`, it crops the image to patch list, calls `extract_feature()` from `torch_utils.py` and return `ImageFeatures` and `ExtractFeaturesReturnMsg`
> * All imports go at the beginning of the file. The only reason I broke this rule is that caffe isn't available in all builds (pip install and clone) so I had to protect those imports.

All imports go at the beginning.

>One more suggestion @qiminchen : can you write a script (pls. put in in the scripts folder), that runs this new feature extractor together with the classifier training on the exported spacer trainingdata and confirms that it get similar (or higher) performance than production. That would function as a regression test basically.

Sure, I will work on it. By the way, I need to add unittest for `EfficientNetExtractor` in `test_extract_features.py` and more specifically I should generate `features_legacy` for `test_regression` right?

qiminchen:
@beijbom hey Oscar, thanks for the review
> * Please re-base to master now that I merged my big PR.

Done
> * If there are utility functions in caffe_utils that you want to use, perhaps move to a shared extract_features_utils file?

- I moved `gray2rgb()`, `crop_patch()` and `crop_simple()` to the shared `extract_features_utils.py`
- `torch_utils.py` only serves as feature extraction

so in `extract_features.py`, it crops the image to patch list, calls `extract_feature()` from `torch_utils.py` and return `ImageFeatures` and `ExtractFeaturesReturnMsg`
> * All imports go at the beginning of the file. The only reason I broke this rule is that caffe isn't available in all builds (pip install and clone) so I had to protect those imports.

All imports go at the beginning.

>One more suggestion @qiminchen : can you write a script (pls. put in in the scripts folder), that runs this new feature extractor together with the classifier training on the exported spacer trainingdata and confirms that it get similar (or higher) performance than production. That would function as a regression test basically.

Sure, I will work on it. By the way, I need to add unittest for `EfficientNetExtractor` in `test_extract_features.py` and more specifically I should generate `features_legacy` for `test_regression` right?

spacer/extract_features.py
```diff
@@ -111,12 +114,50 @@ def feature_dim(self):
...
+        # Set torch parameters
+        torch_params = {'model_type': 'efficientnet',
+                        'model_name': 'efficientnet-b0',
+                        'weights_path': self.modelweighs_path,
+                        'num_class': 1279,
+                        'crop_size': 224,
+                        'batch_size': 10}
+
+        # The imdict data structure needs a label, set to 1, it's not used.
+        rowcollabels = [(row, col, 1) for row, col in rowcols]
```
beijbom, line 136: Why do you use the label here? Maybe it's time to clean that up? It's some old legacy thing I never got around to removing.  Also, please re-factor the caffe extractor to use the same shared utils.

spacer/extract_features_utils.py
```diff
@@ -0,0 +1,64 @@
+import numpy as np
```
beijbom, line 1: Add tests for this file. Most can be moved from `test_caffe_utils.py`.

beijbom:
@qiminchen : sounds good. I added some comments, but let me know when you want me to take a close look again. It's not a bad idea to write a regression test. You can go ahead and upload any test fixtures you want to s3. 

Re. the shared utils, please move them from `caffe_utils` so we don't duplicate code. Also move tests. And please feel free to refactor the caffe extractor also as you are cleaning up and reorganizing the code.

spacer/extract_features.py
```diff
@@ -111,12 +114,50 @@ def feature_dim(self):
...
+
+        # Crop image to patch list
+        im = np.asarray(im)
```
beijbom, line 139: I'd wrap these in a `img_to_patchlist` method. Also, try running your extractor on 
```
image_loc=DataLocation(storage_type='s3',
                                   key='sfq2mr5qbs.jpeg',
                                   bucket_name='spacer-test')
```
and 
```
image_loc=DataLocation(storage_type='s3',
                                   key='kh6dydiix0.jpeg',
                                   bucket_name='spacer-test')
```
These were corner-cases we identified earlier, and for some reason I had to use that do the `im = np.array(pil_im)` conversion twice.

spacer/extract_features.py
qiminchen: label arg removed since it is not needed so I re-factored the Caffe extractor as well

spacer/extract_features_utils.py
qiminchen: added

spacer/extract_features.py
qiminchen: I warped the following to the shared `extract_features_utils.py`, `crop_patch` in `extract_features_utils.py` now works as follow so it can be used by both `VGG16CaffeExtractor` and `EfficientNetExtractor`:
- if gray image, convert to RGB and take the first 3 channels only
- crop and return patch list

In terms of the numpy returning empty array for some images, I found a relevant issue: https://github.com/numpy/numpy/issues/11629 and seems like it's PIL issue
```
_ = np.array(im)  # For some images np.array returns an empty array.
im = np.array(im)  # Running it twice fixes this. Don't ask me why.

if len(im.shape) == 2 or im.shape[2] == 1:
    im = gray2rgb(im)
im = im[:, :, :3]
```

qiminchen:
> One more suggestion @qiminchen : can you write a script (pls. put in in the `scripts` folder), that runs this new feature extractor together with the classifier training on the exported spacer trainingdata and confirms that it get similar (or higher) performance than production. That would function as a regression test basically.

@beijbom Oscar, can you refer me to the files used for extracting features of a source, looks like `scripts/classifier_training_regression.py` downloads the features to local and trains the classifier. What about extracting the features and saving to the temporary directory that you used for VGG16?

For `TestEfficientNetExtractor`, I was able to pass all the cases in local but it fails at Travis CI, it might be because of the `libjpeg` version?
```
  File "/home/travis/build/beijbom/pyspacer/spacer/tests/test_extract_features.py", line 295, in test_regression
    self.assertEqual(pf_new.data, pf_legacy.data)
AssertionError: Lists differ: [1.6237620115280151, -0.1788397878408432, -0.18572[27272 chars]3397] != [1.6237621307373047, -0.17883972823619843, -0.1857[27283 chars]7278]
First differing element 0:
1.6237620115280151
1.6237621307373047
```

beijbom:
> > One more suggestion @qiminchen : can you write a script (pls. put in in the `scripts` folder), that runs this new feature extractor together with the classifier training on the exported spacer trainingdata and confirms that it get similar (or higher) performance than production. That would function as a regression test basically.
> 
> @beijbom Oscar, can you refer me to the files used for extracting features of a source, looks like `scripts/classifier_training_regression.py` downloads the features to local and trains the classifier. 

Yeah, I download them b/c we run 5 epochs, so it's faster to download once and then run locally. You don't have to do that if you don't want to.

> What about extracting the features and saving to the temporary directory that you used for VGG16?

Hmm. I'm not sure what you mean. For extracting features, just compose a `msg  = ExtractFeaturesMsg()` and then call `tasks.extract_features(msg)`. The beauty of the new `DataLocation` design is that `image_loc` and `feature_loc` _can_ point to the same storage system, but they don't have to do so. You can have the task load from the S3 bucket and write to local disk. What ever you prefer. So one way is to download also the image files, and then extract features for all images once they are local. Another way is to point `image_loc` to s3 and `feature_loc` to the local folder. 

> 
> For `TestEfficientNetExtractor`, I was able to pass all the cases in local but it fails at Travis CI, it might be because of the `libjpeg` version?
> 
> ```
> Traceback (most recent call last):
>   File "/home/travis/build/beijbom/pyspacer/spacer/tests/test_extract_features.py", line 295, in test_regression
>     self.assertEqual(pf_new.data, pf_legacy.data)
> AssertionError: Lists differ: [1.6237620115280151, -0.1788397878408432, -0.18572[27272 chars]3397] != [1.6237621307373047, -0.17883972823619843, -0.1857[27283 chars]7278]
> First differing element 0:
> 1.6237620115280151
> 1.6237621307373047
> ```

I don't think this has to do with `libjpeg`. When you compare float32 vectors you can't expect identical results. Instead use a test that is robust to float precision errors. For example, you can borrow from what I did here:
https://github.com/beijbom/pyspacer/blob/master/spacer/tests/test_beta_regression.py#L141

beijbom:
One thing to note about Travis CI. Some unit-tests that involved fetching stuff from S3 do fail intermittently. So if you see fails related to that ping me and I can restart that job on Travis. In the long term we should probably move to using local test fixtures.

qiminchen:
@beijbom hey Oscar, the script that runs the new feature extractor together with the classifier training on the exported spacer trainingdata is ready and I tested it on `s294`, the performance is better than the production as we're using `EfficientNet-b0`. The new repo is a legend!
```
-> Downloading data for source id: 294.
-> Downloading 201 metadata...
-> Assembling train and val data for source id: 294
-> Training...
-> Trainset: 157, valset: 18 images
-> Using 25 images per mini-batch and 7 mini-batches per epoch
-> Trainset: 20, valset: 9, common: 8 labels
-> Loading reference data.
-> Online training...
-> Epoch 0, acc: 0.8368991386496248
-> Epoch 1, acc: 0.8388441233676021
-> Epoch 2, acc: 0.8399555432064463
-> Epoch 3, acc: 0.8396776882467352
-> Epoch 4, acc: 0.8407891080855793
-> Epoch 5, acc: 0.8405112531258683
-> Epoch 6, acc: 0.8421783828841345
-> Epoch 7, acc: 0.8421783828841345
-> Epoch 8, acc: 0.8419005279244235
-> Epoch 9, acc: 0.8424562378438455
-> Calibrating.
-> Re-trained LTER Back Reef (294). Old acc: 80.0, new acc: 86.8
```
I tried to make the script more flexible by:
```
def main(extractor_name, source_id, n_epochs):
    extractor = ExtractFeatures(extractor_name)
    extractor(source_id)
    classifier = TrainClassifier()
    classifier(source_id, n_epochs)

if __name__ == '__main__':
    fire.Fire(main)
```
then I ran into some relative import issue and haven't found a better way to solve it, `e.g.`
```
  File "scripts/features_extracting_training.py", line 16, in <module>
    from spacer import config
ModuleNotFoundError: No module named 'spacer'
``` 
Do you have any ideas? Other than that, this PR should be ready for review.

spacer/extract_features_utils.py
```diff
@@ -0,0 +1,71 @@
...
+    w, h = im.shape
+    ret = np.empty((w, h, 3), dtype=np.uint8)
+    ret[:, :, 0] = im
+    ret[:, :, 1] = im
+    ret[:, :, 2] = im
+
+    return ret
+
+
+def crop_patch(im, rowcols, crop_size, scale, offsets=None):
```
beijbom, line 19: since we only support `scale=1`, go ahead and remove this parameter. We can always add it in later if needed.

spacer/models/effcientnet_utils.py
```diff
@@ -0,0 +1,298 @@
+"""
```
beijbom, line 1: I know these files are borrowed from another repo, but can you do a PEP8 clean-up anyways? Check line length etc.? I have been using 79 character PEP8 line length.

spacer/extract_features_utils.py
beijbom, line 19: Also: maybe add type hints here? I realized I hadn't done so originally, but since this is now a shared util, seems worth it.

beijbom: ... we really should have them for all methods. I just got lazy in the end.

spacer/config.py
```diff
@@ -121,6 +121,8 @@ def get_local_model_path():
 
 LOCAL_FIXTURE_DIR = os.path.join(os.path.dirname(
     os.path.abspath(__file__)), 'tests', 'fixtures')
+LOCAL_FEATURE_DIR = os.path.join(os.path.dirname(
```
beijbom, line 124: This is only used in the scripts, right? If so, let's remove it from here, and make it an input to the script CLI. The scripts are not part of the PIP package so they should be as stand-alone as possible.

spacer/tests/test_extract_features.py
```diff
@@ -177,5 +178,133 @@ def test_regression(self):
...
+                                       key='08bfc10v7t.png.effnet.featurevector',
+                                       bucket_name='spacer-test')
+
+        ext = feature_extractor_factory(msg.feature_extractor_name)
+
+        img = load_image(msg.image_loc)
+        features_new, _ = ext(img, msg.rowcols)
+        features_legacy = ImageFeatures.load(legacy_feat_loc)
+
+        self.assertFalse(features_legacy.valid_rowcol)
```
beijbom, line 294: This is confusing. Why is `valid_rowcol = False`? Didn't you extract these features using this code-base? If so, the ImageFeatures dataclass should be stored with valid rowcols. The legacy thing is just for the old extractor deployed on CoralNet..

spacer/tests/test_extract_features.py
```diff
@@ -177,5 +178,133 @@ def test_regression(self):
...
+        self.assertEqual(features_legacy.npoints, len(rowcols))
+        self.assertEqual(features_legacy.feature_dim, 1280)
+
+        self.assertTrue(features_new.valid_rowcol)
+        self.assertEqual(features_new.npoints, len(rowcols))
+        self.assertEqual(features_new.feature_dim, 1280)
+
+        for pf_new, pf_legacy in zip(features_new.point_features,
+                                     features_legacy.point_features):
+            self.assertTrue(np.allclose(pf_legacy.data, pf_new.data, rtol=1e-4))
```
beijbom, line 304: Also, you are right. This passes fine on my dev machine, but not on Travis CI. Maybe this was due to that `libjpeg` after all? Hmm. I suppose you could change to using `png` format instead.

beijbom: ... actually, it passes fine now. Nevermind -- let's leave it as is for now.

beijbom: ... actually, it does not pass. Can you try using `png`?

scripts/features_extracting_training.py
```diff
@@ -0,0 +1,154 @@
...
+        all_keys = bucket.list(prefix='{}/s{}/images'.format(export_name, source_id))
+        img_keys = [key.name for key in all_keys if key.name.endswith('jpg')]
+        ann_keys = [Key(bucket=bucket, name=img.replace('jpg', 'anns.json')) for img in img_keys]
+
+        feats_local_path = os.path.join(config.LOCAL_FEATURE_DIR, 's{}'.format(source_id), 'images')
+        if not os.path.isdir(feats_local_path):
+            os.system('mkdir -p ' + feats_local_path)
+
+        print("-> Extracting features...")
+        for idx, im_key in enumerate(img_keys):
```
beijbom, line 45: Here, I suggest checking if the target `feature_loc` already exists on disk. If so, skip that iteration. This allows users to re-start the script if it hangs (mine did).

spacer/models/effcientnet_utils.py
qiminchen: sure, I've been using 120 but I will change it to 79 for consistency

spacer/extract_features_utils.py
qiminchen: done

spacer/config.py
qiminchen: make sense, done

spacer/tests/test_extract_features.py
qiminchen: no, I didn't use this code-base to extract the legacy features, instead, I used the one in [CoralNet](https://github.com/qiminchen/CoralNet/blob/f0c5985cdc812ec8e42a880035fa5961a8f894d6/extract.py#L82) repo, I was thinking if we use the same code to extract features, the features are guaranteed to be the same without concerning `e.g. libjpeg` issue, though we know the new extractor is correct as the regression test work so well. what do you think?

spacer/extract_features_utils.py
qiminchen: hah yea sure, I will go through all utils files and add type hints

scripts/features_extracting_training.py
qiminchen: updated

spacer/tests/test_extract_features.py
beijbom: I see. So the legacy is from your own codebase. I like that. So perhaps we just try with `pngs` to rule out the libjpeg issue during the tests?

spacer/tests/test_extract_features.py
qiminchen: actually the images I used for generating the legacy features are already `png`, and the original image for production testing is also `png` so technically it should work. Weird...

qiminchen:
@beijbom hey Oscar, here are some updates on the progress:
> * Can you go through and add Typing hints to all utils? I was a bit lazy previously, but we might as well do it.

I should've completed adding type hints to files that needed. Please let me know if I miss any
> * Code coverage for the files in `/models` folder is not 100%. Can you add test to cover those lines?

Added, should be coveraged now, already tested in local

> * For the scripts, I think there are some duplicated code between `features_extracting_training.py` and `classifier_training_regression.py`. Can you move the shared code to a new `scripts.utils.py`?

Added. I moved the `build_traindata()` and `start_training()` to the `scripts_utils.py` and cleaned up both `features_extracting_training.py` and `classifier_training_regression.py`. 

The reason I didn't move `_cache()` to `scripts_utils.py` is that `classifier_training_regression.py` downloads all the files to local (`features.json`, `anns.json` and `image.meta.json`) while `features_extracting_training.py` only downloads `anns.json` and `image.meta.json` as `features` are downloaded when being extracted.

> * As per your question. Hmm. That should work fine. Can you push that change so I can see if I can replicate the problem? My guess is that you may have to add spacer to your PYTHONPATH.

Thanks for the help, I solved this issue by adding the `spacer` to `$PYTHONPATH`, so `config.LOCAL_FEATURE_DIR` goes to the CLI, you can use 

```$ python scripts/features_extracting_training.py efficientnet_b0_ver1 294 10 \ /path/to/local/path/for/saving/features```

where `efficientnet_b0_ver1 294` is extractor_name, `294` is source_id and `10` is the number of training epochs, to run the `features_extracting_training.py`

> * actually, it does not pass. Can you try using `png`?

Actually the images I used for generating the legacy features are already `png`, and the original image for production testing is also `png` so technically it should work. Weird... still debugging...

spacer/tests/test_extract_features.py
qiminchen: This is so weird as I checked the array values of every step using my codebase and spacer and they are all the same: original images value, cropped patches value, before and after pre-processing value, all with `png` format. And I did turn off the BN and dropout otherwise the results will be changed every time the extractor is called. The thing is it works perfectly in local...

beijbom: I think I fixed it. Turns out you were setting the wrong parameter of `np.allclose`. This is the formula: `absolute(arr1 - arr2) <= (atol + rtol * absolute(arr2))`. You were setting `rtol` < 1 which made the criterion even more strict. I set `atol=1e-4` and that did the trick.

beijbom: (I'm still seeing problems related to lag connecting to S3 from Travis, but that is a different story, usually works after a restart).

qiminchen: Great!! BTW EfficientNet-b0 training on dataset V3 will be finished in a few hours, I will use this legendary for evaluation.

beijbom:
@qiminchen : Great job pushing this PR through. Feel free to merge this to master when you are ready.

StephenChan:
Nice work on this!

PR #18, opened 2020-05-01
minor update

qiminchen:
1. remove self.maxDiff otherwise all the difference will be logged
2. type `Roe` -> `Row`
3. `im_height` should compare with `row`, not `col`

### DIFF
--- a/scripts/features_extracting_training.py
+++ b/scripts/features_extracting_training.py
@@ -64,7 +64,7 @@ class ExtractFeatures:
                 anns = json.load(BytesIO(
                     ann_keys[idx].get_contents_as_string()
                 ))
-                rowcols = [(ann['row'], ann['col']) for ann in anns]
+                rowcols = [(ann['row']-1, ann['col']-1) for ann in anns]
 
                 img_loc = DataLocation(storage_type='s3',
                                        key=im_key,
--- a/scripts/scripts_utils.py
+++ b/scripts/scripts_utils.py
@@ -31,7 +31,7 @@ def build_traindata(image_root: str) -> Tuple[ImageLabels, ImageLabels]:
             labels = val_labels
 
         labels.data[features_file] = [
-            (ann['row'], ann['col'], ann['label']) for ann in anns
+            (ann['row']-1, ann['col']-1, ann['label']) for ann in anns
         ]
     return train_labels, val_labels
 
--- a/spacer/task_utils.py
+++ b/spacer/task_utils.py
@@ -22,7 +22,7 @@ def check_rowcols(rowcols: List[Tuple[int, int]],
             "Columns must be non-negative. Given was: [{}]".format(col)
 
         assert row < im_height, \
-            "Roe {} outside image with nrows: {}".format(col, im_height)
+            "Row {} outside image with nrows: {}".format(row, im_height)
 
         assert col < im_width, \
             "Column {} outside image with ncols: {}".format(col, im_width)
--- a/spacer/tests/test_extract_features.py
+++ b/spacer/tests/test_extract_features.py
@@ -264,7 +264,6 @@ class TestEfficientNetExtractor(unittest.TestCase):
         self.assertEqual(len(features.point_features[0].data), 1280)
 
     def test_regression(self):
-        self.maxDiff = None
         rowcols = [(20, 265),
                    (76, 295),
                    (59, 274),


qiminchen:
@beijbom hi Oscar, some patches happen to be annotated in the edge of the image, I'm not sure if it is normal, should we change 

`assert row < im_height, assert col < im_width` -> `assert row <= im_height, assert col <= im_width`?
```
  File "scripts/features_extracting_training.py", line 165, in <module>
    fire.Fire(main)
  File "/home/qimin/anaconda3/envs/pyspacer/lib/python3.7/site-packages/fire/core.py", line 138, in Fire
    component_trace = _Fire(component, args, parsed_flag_args, context, name)
  File "/home/qimin/anaconda3/envs/pyspacer/lib/python3.7/site-packages/fire/core.py", line 468, in _Fire
    target=component.__name__)
  File "/home/qimin/anaconda3/envs/pyspacer/lib/python3.7/site-packages/fire/core.py", line 672, in _CallAndUpdateTrace
    component = fn(*varargs, **kwargs)
  File "scripts/features_extracting_training.py", line 159, in main
    extractor(source_id)
  File "scripts/features_extracting_training.py", line 83, in __call__
    _ = extract_features(msg)
  File "/home/qimin/Projects/pyspacer/spacer/tasks.py", line 26, in extract_features
    check_rowcols(msg.rowcols, img)
  File "/home/qimin/Projects/pyspacer/spacer/task_utils.py", line 25, in check_rowcols
    "Row {} outside image with nrows: {}".format(row, im_height)
AssertionError: Row 2136 outside image with nrows: 2136
```

beijbom:
Interesting. I suppose it depends on whether we use zero-indexing one one-indexing. I have been assuming zero-indexing in which case `Row 2136 outside image with nrows: 2136` is correct. Where did you get these data from? The coralnet beta-export? 
@StephenChan : are we using zero-indexing in CoralNet? Do we allow points outside the image?

qiminchen:
@beijbom yes, the data is from coralnet beta-export.

BTW, I'm adding function to export the confusion matrix on val set when the training is finished

beijbom:
> @beijbom yes, the data is from coralnet beta-export.

Interesting. Can you share the source and image id? Let's pull it up on CoralNet and see what it looks like. It's not really a  problem, but I'm a bit confused as to why it happened. I can look at the coralnet source code also, should be somewhere in https://github.com/beijbom/coralnet/blob/master/project/images/models.py.

qiminchen:
> Interesting. Can you share the source and image id? Let's pull it up on CoralNet and see what it looks like. It's not really a problem, but I'm a bit confused as to why it happened. I can look at the coralnet source code also, should be somewhere in https://github.com/beijbom/coralnet/blob/master/project/images/models.py.

s1420: i1103958.jpg
s601: i475249.jpg

yea I don't think this is a big problem and patches located at the edge only compose of a small portion of the source.

beijbom:
> > Interesting. Can you share the source and image id? Let's pull it up on CoralNet and see what it looks like. It's not really a problem, but I'm a bit confused as to why it happened. I can look at the coralnet source code also, should be somewhere in https://github.com/beijbom/coralnet/blob/master/project/images/models.py.
> 
> s1420: i1103958.jpg
> s601: i475249.jpg
> 
> yea I don't think this is a big problem and patches located at the edge only compose of a small portion of the source.

Ok, yeah, point 19 is actually outside the image: https://coralnet.ucsd.edu/image/1103958/annotation/tool/. Thanks for catching this.

If CoralNet indeed is 1-indexed, we could just subtract 1 from each row&col before submitting to spacer. If not, and e.g. both row=0 and row=nbr_rows is valid, then there is an error in CoralNet. So the question is what do we do then?

1) Change the code on the spacer side to allow row/col locations _outside_ the image. In this case we need to think about how far outside we would allow. 
2) Do not change the code on the spacer side. In this situation the client code should handle it. In other words, check the size of the image and move the row/col locations if needed.

beijbom:
@qiminchen : It looks like CoralNet is 1-indexed. See this line:
https://github.com/beijbom/coralnet/blob/master/project/annotations/model_utils.py#L96

So you need to subtract `1` from each row, col before submitting to spacer.

@StephenChan : does that sound right to you?

StephenChan:
Yeah, I was just about to chime in. CoralNet rows/cols start at 1.

qiminchen:
subtracting `1` from each row, col before submitting to spacer sounds good, allowing `row/col` locations outside the image probably won't be a good idea as annotation outside the image shouldn't be considered as a valid annotation.

qiminchen:
@beijbom tested on `s601` and `s1420` again and works well, should be fine now, ready for review

PR #17, opened 2020-04-29
AWS sqs deploy scripts.

beijbom:
This PR will has scripts to test the new compute cluster on AWS. These scripts all rely on the `spacer_test_jobs` and `spacer_test_results` queues. These queues are isolated from the production queues but share the same compute instances, so we can test stuff like memory, runtime, etc.

Also note that this cluster is a brand new one, totally isolated from the cluster used for the current production environment. I will leave both up and running until we have completed the transition.

Action points:
- [x] Setup new cluster on AWS.
- [x] Add rules for scaling up and down.
- [x] Move to standard queues for production and test.
- [x] Run scaling shakeouts.
- [x] Clean up scripts on spacer side.
- [x] Enable ECS to run jobs on both test and production queue.
- [x] Update docker build after https://github.com/beijbom/pyspacer/pull/15 is done.
- [x] Run memory tests for large images.
- [x] Run memory tests for large number of row/col locations.
- [x] Add memory and image size limits and write tests.

Summary of changes:
- Refactored and added scripts to `scripts` folder. New scripts are in `aws` subfolder.
- Added asserts on image size and number of pixels. These were set based on experiments shown below. See this commit for details: https://github.com/beijbom/pyspacer/pull/17/commits/9bc366d69a5494cd2e50f82917da2b0956d7f063
- Refactor of `extract_feature` in **spacer/torch_utils.py** to reduce memory footprint. See commit: https://github.com/beijbom/pyspacer/pull/17/commits/318dddbb129284ed072f0ffc49b2d559e47b6791.
- PEP8 and imports cleanup throughout.
- All tests pass on Docker.

PR #24, opened 2020-05-09
numpy vs pil simple memory usage

qiminchen:
This is a simple memory usage comparison of `Numpy` and `PIL` mentioned at https://github.com/beijbom/pyspacer/pull/17#discussion_r420909318

### DIFF
--- a/spacer/extract_features_utils.py
+++ b/spacer/extract_features_utils.py
@@ -60,3 +60,20 @@ def crop_simple(im: np.ndarray,
     left = int(center[1] - crop_size / 2)
 
     return im[upper: upper + crop_size, left: left + crop_size, :]
+
+
+def crop_patches_pil(im: Image,
+                     rowcols: List[Tuple[int, int]],
+                     crop_size: int):
+    pad = crop_size
+    image = Image.fromarray(
+        np.pad(im.convert('RGB'),
+               ((pad, pad), (pad, pad), (0, 0)), mode='reflect'),
+        'RGB'
+    )
+
+    return [image.crop((pad + col - crop_size / 2,
+                        pad + row - crop_size / 2,
+                        pad + col + crop_size / 2,
+                        pad + row + crop_size / 2))
+            for row, col in rowcols]
--- a/spacer/extract_features.py
+++ b/spacer/extract_features.py
@@ -12,7 +12,7 @@ from PIL import Image
 
 from spacer import config
 from spacer.data_classes import PointFeatures, ImageFeatures
-from spacer.extract_features_utils import crop_patches
+from spacer.extract_features_utils import crop_patches, crop_patches_pil
 from spacer.messages import ExtractFeaturesReturnMsg
 from spacer.storage import download_model
 from spacer.torch_utils import extract_feature
@@ -120,7 +120,7 @@ class EfficientNetExtractor(FeatureExtractor):
         self.modelweighs_path, self.model_was_cashed = download_model(
             'efficientnetb0_5eps_best.pt')
 
-    def __call__(self, im, rowcols):
+    def __call__(self, im, rowcols, types='numpy'):
 
         start_time = time.time()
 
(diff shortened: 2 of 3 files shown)


qiminchen:
@beijbom hey Oscar, as you mentioned at https://github.com/beijbom/pyspacer/pull/17#discussion_r420909318, one way to save more memory is to crop the image first using `Image.crop()` from `PIL` and convert the patches to `Numpy` after. So I did a very simple experiment to trace memory usage. Method for monitoring the memory usage adopted from [Option 2: tracemalloc](https://medium.com/survata-engineering-blog/monitoring-memory-usage-of-a-running-python-program-49f027e3d1ba) and refer to [tracemalloc](https://docs.python.org/3/library/tracemalloc.html#tracemalloc.get_traced_memory) for more details.

Basically, `crop_patches_pil()` in `extract_features_utils.py` return a list of cropped PIL Image, `if...else...` in `extract_features.py` just make the comparison convenient. `TestMemoryUsage` in `test_extract_features.py` simply test the memory usage of extraction block. These auxiliary functions will be removed if we decide to merge this PR.

`tracemalloc` basically tracks the individual memory blocks allocated by the Python interpreter when `__call__(self, im, rowcols)` in `EfficientNetExtractor()` is completed so I assume this is equivalent to the memory usage measurement.

Result on `08bfc10v7t.png` with 5 annotations. Looks like cropping the image first using `Image.crop()` from `PIL` and convert the patches to `Numpy` after would save more memory than converting to numpy array at the beginning.
```
-> Initializing EfficientNetExtractor
Numpy: Current memory usage is 0.464665MB; Peak was 4.160963MB
-> Initializing EfficientNetExtractor
PIL: Current memory usage is 0.46915MB; Peak was 2.939177MB
```

spacer/extract_features_utils.py
```diff
@@ -60,3 +60,20 @@ def crop_simple(im: np.ndarray,
...
+                     rowcols: List[Tuple[int, int]],
+                     crop_size: int):
+    pad = crop_size
+    image = Image.fromarray(
+        np.pad(im.convert('RGB'),
+               ((pad, pad), (pad, pad), (0, 0)), mode='reflect'),
+        'RGB'
+    )
+
+    return [image.crop((pad + col - crop_size / 2,
```
beijbom, line 75: Image.crop returns Image objects, so you'd need to convert to arrays here to match with the previous method. https://pillow.readthedocs.io/en/3.1.x/reference/Image.html#PIL.Image.Image.crop

beijbom:
hey @qiminchen . Thanks for looking into this. It's interesting it saves so much memory since you are still converting it to an array in order to do the np.pad call. How much does this save on a 100 mega pixel image?

I'm inclined to stay with the old code since we know that it works (don't fix what ain't broken...). Also, I have had some bad experience with the PIL.convert('RGB') step -- just seem to cause trouble for some of the more unusual image formats. 

But let's keep an eye on this fix if we run into memory issues down the road.

spacer/extract_features_utils.py
qiminchen: if we convert to arrays here, wouldn't it be the same as the original `crop_patches`?

In Caffe, we have `Transformer` that can process `Image` object https://github.com/beijbom/pyspacer/blob/b106de785e21e14cfbac03878ff85f1b48b80847/spacer/caffe_utils.py#L26-L34
In Pytorch, we have `transformation` that can process `Image` object as well
https://github.com/beijbom/pyspacer/blob/b106de785e21e14cfbac03878ff85f1b48b80847/spacer/torch_utils.py#L15-L24

qiminchen:
@beijbom 
> How much does this save on a 100 mega pixel image?

I will take a look at this after.
> I'm inclined to stay with the old code since we know that it works (don't fix what ain't broken...). Also, I have had some bad experience with the PIL.convert('RGB') step -- just seem to cause trouble for some of the more unusual image formats.

I agree.

beijbom:
@qiminchen : I'm closing this PR for now. We can revisit later.

PR #26, opened 2020-06-06
Integration shakeout

beijbom:
- Fixes a bug in the `train_classifier` task.
- Allows filesystem storage to create folders.
- Adds more tests and asserts.

### DIFF
--- a/spacer/tasks.py
+++ b/spacer/tasks.py
@@ -3,6 +3,7 @@ Defines the highest level methods for completing tasks.
 """
 import time
 import logging
+import traceback
 
 from spacer import config
 from spacer.data_classes import ImageLabels, ImageFeatures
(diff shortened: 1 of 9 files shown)


spacer/tasks.py
```diff
@@ -49,7 +50,7 @@ def train_classifier(msg: TrainClassifierMsg) -> TrainClassifierReturnMsg:
     # Do the actual training
     clf, val_results, return_message = trainer(
         ImageLabels.load(msg.traindata_loc),
-        ImageLabels.load(msg.traindata_loc),
+        ImageLabels.load(msg.valdata_loc),
```
beijbom, line 53: @qiminchen : I'm integrating spacer in CoralNet and stumbled across this rather nasty bug. Wanted to let you know in case you are using this for your work.

qiminchen: thanks for letting me this, all good

PR #27, opened 2020-06-09
sumcheck and replace with new model weight

qiminchen:
- new model weight uploaded to spacer-tools with a name of `efficientnet_b0_ver1.pt`
- `num_class` changed to `1275`
- add `sha256` sumcheck to verify downloading the weights correctly
- since using the new model weight, the legacy test feature is regenerated.

### DIFF
--- a/spacer/caffe_utils.py
+++ b/spacer/caffe_utils.py
@@ -9,7 +9,10 @@ from functools import lru_cache
 from typing import List, Any, Tuple
 
 import caffe
+import hashlib
 import numpy as np
+
+from spacer import config
 
 
 class Transformer:
@@ -96,6 +99,11 @@ def load_net(modeldef_path: str,
     :param modelweighs_path: pretrained weights path.
     :return: pretrained model.
     """
+    # To verify that the correct weight is loaded
+    with open(modelweighs_path, 'rb') as fp:
+        sha256 = hashlib.sha256(fp.read()).hexdigest()
+    assert sha256 == config.MODEL_WEIGHTS_SHA['vgg16']
+
     return caffe.Net(modeldef_path, modelweighs_path, caffe.TEST)
 
 
--- a/spacer/caffe_utils.py
+++ b/spacer/caffe_utils.py
@@ -4,6 +4,7 @@ simplicity. Since support for Caffe will be deprecate,
 these are only lightly cleaned up from their original state.
 """
 
+import logging
 from copy import copy
 from functools import lru_cache
 from typing import List, Any, Tuple
@@ -105,8 +106,9 @@ def load_net(modeldef_path: str,
     with open(modelweighs_path, 'rb') as fp:
         sha256 = hashlib.sha256(fp.read()).hexdigest()
     assert sha256 == config.MODEL_WEIGHTS_SHA['vgg16']
-    print("==> time spent on checking sha: {}".format(time.time() - start))
-
+    logging.debug("-> Time spent on checking SHA: {}".format(
+        time.time() - start
+    ))
     return caffe.Net(modeldef_path, modelweighs_path, caffe.TEST)
 
 
@@ -128,9 +130,7 @@ def classify_from_patchlist(patchlist: List,
     """
     # Setup caffe
     caffe.set_mode_cpu()
-    start = time.time()
     net = load_net(modeldef_path, modelweighs_path)
-    print("==> time spent in total: {}".format(time.time() - start))
 
     # Classify
     transformer = Transformer(pyparams['im_mean'])
--- a/spacer/caffe_utils.py
+++ b/spacer/caffe_utils.py
@@ -9,6 +9,7 @@ from functools import lru_cache
 from typing import List, Any, Tuple
 
 import caffe
+import time
 import hashlib
 import numpy as np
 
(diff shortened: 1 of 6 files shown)


qiminchen:
@beijbom I uploaded the newly trained model weight to the spacer-tools as `efficientnet_b0_ver1.pt` and update all the places it appears in the codebase. For the sumcheck, I'm not sure if I add it correctly but pretty sure it downloads and uses the correct weight. `c3dc6d304179c6729c0a0b3d4e60c728bdcf0d82687deeba54af71827467204c` is generated in my local.

spacer/torch_utils.py
```diff
@@ -33,6 +34,10 @@ def load_weights(model: Any,
     :param modelweighs_path: pretrained model weight from new CoralNet
     :return: well trained model
     """
+    with open(modelweighs_path, 'rb') as fp:
+        sha256 = hashlib.sha256(fp.read()).hexdigest()
+    assert sha256 == \
+           'c3dc6d304179c6729c0a0b3d4e60c728bdcf0d82687deeba54af71827467204c'
```
beijbom, line 40: `load_weights`  is intended to support any EfficientNet implementation, right? So the sha needs to be specific to this `feature_extractor_name`. I suggest you a dict in the main config file indexed on the `feature_extractor_name`. Can you add the same for the caffe extractor pls.?

Also, it looks like you are loading the weights twice. Once to calculate the sha, and then again below. Can you try to read it just once?

qiminchen: sure, i can add the same for the Caffe, do you want me to move the sha to the config?

> Also, it looks like you are loading the weights twice. Once to calculate the sha, and then again below. Can you try to read it just once?

ok, working on it

beijbom: Yeah. Each feature_extractor should have a weight check-sum to make sure we are loading the weights we think we are. These should be stored in the main config file.

qiminchen: > `load_weights` is intended to support any EfficientNet implementation, right? So the sha needs to be specific to this feature_extractor_name

Yes, `load_weights` can support any EfficientNet implementation and actually `class EfficientNetExtractor(FeatureExtractor)` support any EfficientNet so I dictionarize the sha indexed on the `model_name` instead of `feature_extractor_name`, what do you think?

Checksum added on the Caffe and modify the checksum in Pytorch to load the weights only once.

spacer/caffe_utils.py
```diff
@@ -96,6 +99,11 @@ def load_net(modeldef_path: str,
     :param modelweighs_path: pretrained weights path.
     :return: pretrained model.
     """
+    # To verify that the correct weight is loaded
+    with open(modelweighs_path, 'rb') as fp:
```
beijbom, line 103: Hmm. So in this case we are forced to load it twice, it seems, since the loading happens in the `caffe.Net()` call. Can you measure how long this extra step takes?

qiminchen:
```
./usr/lib/python3.6/importlib/_bootstrap.py:219: RuntimeWarning: numpy.ufunc size changed, may indicate binary incompatibility. Expected 192 from C header, got 216 from PyObject
  return f(*args, **kwds)
==> time spent on checking sha: 0.44178128242492676
[libprotobuf WARNING google/protobuf/io/coded_stream.cc:604] Reading dangerously large protocol message.  If the message turns out to be larger than 2147483647 bytes, parsing will be halted for security reasons.  To increase the limit (or to disable these warnings), see CodedInputStream::SetTotalBytesLimit() in google/protobuf/io/coded_stream.h.
[libprotobuf WARNING google/protobuf/io/coded_stream.cc:81] The total number of bytes read was 552712164
==> time spent in total: 0.9515583515167236
.==> time spent in total: 8.821487426757812e-06
.==> time spent on checking sha: 0.44198107719421387
[libprotobuf WARNING google/protobuf/io/coded_stream.cc:604] Reading dangerously large protocol message.  If the message turns out to be larger than 2147483647 bytes, parsing will be halted for security reasons.  To increase the limit (or to disable these warnings), see CodedInputStream::SetTotalBytesLimit() in google/protobuf/io/coded_stream.h.
[libprotobuf WARNING google/protobuf/io/coded_stream.cc:81] The total number of bytes read was 552712164
.==> time spent in total: 1.6689300537109375e-06
..............==> time spent in total: 5.9604644775390625e-06
.==> time spent in total: 5.7220458984375e-06
..==> time spent in total: 4.291534423828125e-06
.==> time spent in total: 4.0531158447265625e-06
```
This is the measurement on the docker, roughly extra `0.44178128242492s` is spent on checking the `sha256`, check the code in `caffe_utils.py`

qiminchen: @beijbom

beijbom: Ok. I think that is fine.

spacer/caffe_utils.py
```diff
@@ -96,6 +100,13 @@ def load_net(modeldef_path: str,
     :param modelweighs_path: pretrained weights path.
     :return: pretrained model.
     """
+    # To verify that the correct weight is loaded
+    start = time.time()
+    with open(modelweighs_path, 'rb') as fp:
+        sha256 = hashlib.sha256(fp.read()).hexdigest()
+    assert sha256 == config.MODEL_WEIGHTS_SHA['vgg16']
+    print("==> time spent on checking sha: {}".format(time.time() - start))
```
beijbom, line 108: Change to a `logger.debug` message

spacer/caffe_utils.py
```diff
@@ -117,7 +128,9 @@ def classify_from_patchlist(patchlist: List,
     """
     # Setup caffe
     caffe.set_mode_cpu()
+    start = time.time()
     net = load_net(modeldef_path, modelweighs_path)
+    print("==> time spent in total: {}".format(time.time() - start))
```
beijbom, line 133: `logger.debug`

beijbom:
@qiminchen : please finish last items and remove "work in progress" label once ready for final review.

PR #28, opened 2020-09-08
New EfficientNet-B0 / MLP classifier deployment

qiminchen:
~~Upload new EfficientNet-bo well-trained weights~~
- [x] Add MLP classifier
- [x] Unit test for MLP classifier hybrid mode
- [x] Unit tests all passed on local
- [x] Unit tests all passed on Docker


qiminchen:
hey @beijbom, I added the MLP classifier by adding one more argument `clf_type: str # Name of the classifier to use`, tested on both local and docker, passed most test cases except this one. I don't think this failure is caused by adding the MLP. Do you have any ideas?
```
======================================================================
FAIL: test_img_classify_bad_url (spacer.tests.test_mailman.TestProcessJobErrorHandling)
----------------------------------------------------------------------
  File "/workspace/spacer/spacer/tests/test_mailman.py", line 62, in test_img_classify_bad_url
    self.assertTrue('URLError' in return_msg.error_message)
AssertionError: False is not true
```
I set the default `clf_type='MLP` and ran the 
`python scripts/regression/efficientnet_extractor.py efficientnet_b0_ver1 294 10 MLP /path/to/features`
to test both EfficientNet-b0 feature extraction and MLP classifier training, it worked pretty well. You could also test on your local. Changes are ready for review.

beijbom:
Hey. Looks like some other tests fail on Travis: https://travis-ci.org/github/beijbom/pyspacer/jobs/725454742

beijbom:
Re. the error. I'm not sure. Can you print the error message (`return_msg.error_message`)? Perhaps the formatting changed slightly so it doesn't match what I check against?

qiminchen:
> Re. the error. I'm not sure. Can you print the error message (`return_msg.error_message`)? Perhaps the formatting changed slightly so it doesn't match what I check against?

I know what the error is, working on it

beijbom:
Btw. I updated CI to use travis-ci.com. CI callbacks seem to be back up again.

qiminchen:
@beijbom It should work well now

spacer/tests/test_mailman.py
```diff
@@ -100,25 +100,27 @@ def test_img_classify_classifier_key(self):
...
-                         previous_model_locs=[
-                             DataLocation(storage_type='memory',
-                                          key='my_previous_model')
-                         ],
-                         model_loc=DataLocation(storage_type='memory',
-                                                key='model'),
-                         valresult_loc=DataLocation(storage_type='memory',
-                                                    key='val_res'))])
+        for clf_type in ['LR', 'MLP']:
+            msg = JobMsg(task_name='train_classifier',
```
beijbom, line 104: tests needs to be indented also

spacer/train_utils.py
```diff
@@ -60,8 +62,15 @@ def train(labels: ImageLabels,
...
     # Initialize classifier and ref set accuracy list
-    logging.info("-> Online training...")
-    clf = SGDClassifier(loss='log', average=True, random_state=0)
+    logging.info("-> Online training using {}...".format(clf_type))
+    if clf_type == 'MLP':
+        if len(train_set) * labels.samples_per_image >= 50000:
+            hls, lr = (200, 100), 1e-4
+        else:
+            hls, lr = (100,), 1e-3
+        clf = MLPClassifier(hidden_layer_sizes=hls, learning_rate_init=lr)
```
qiminchen, line 71: @beijbom hybrid mode setup here

spacer/tests/test_mailman.py
qiminchen: fixed

spacer/tests/test_train_utils.py
```diff
@@ -90,34 +117,38 @@ def test_too_few_classes(self):
...
+                              num_epochs, clf_type)
 
 
 class TestEvaluateClassifier(unittest.TestCase):
 
     def test_simple(self):
         feature_loc = DataLocation(storage_type='memory', key='')
         train_data = make_random_data(10, [1, 2], 4, 5, feature_loc)
-        clf, _ = train(train_data, feature_loc, 1)
+        for clf_type in ['LR', 'MLP']:
```
beijbom, line 130: Better to iterate over `config.CLASSIFIER_TYPES` since 1) this test will test new stuff you add in the future. 2) you help other coders realize this list is defined in the config.

beijbom:
@qiminchen : I took a look -- this looks nice in general. I ran
```
% python scripts/regression/train_classifier.py train 1498 /data/tmp
-> Downloading data for source id: 1498.
-> Downloading 804 metadata and feature files...
100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 804/804 [03:08<00:00,  4.26it/s]
-> Assembling train and val data for source id: 1498
-> Training...
-> Re-trained SpatSurvey (1498). Old acc: 76.0, new acc: 72.5
```

which gave lower performance than the previous setup. Do you mind running a few sources and pasting the results here? I just like to confirm that this is an outlier.

Also: it seems we should be able to clean up the `regression` folder a bit. For example, there are two private methods for caching local, one in `efficientnet_extractor.py` and one in `train_classifier.py`. I know they do slightly different things, but can you see if it can be moved to a shared `utils.py`?

spacer/tests/test_train_utils.py
qiminchen: you're right, working on this

qiminchen:
> @qiminchen : I took a look -- this looks nice in general. I ran
> 
> ```
> % python scripts/regression/train_classifier.py train 1498 /data/tmp
> -> Downloading data for source id: 1498.
> -> Downloading 804 metadata and feature files...
> 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 804/804 [03:08<00:00,  4.26it/s]
> -> Assembling train and val data for source id: 1498
> -> Training...
> -> Re-trained SpatSurvey (1498). Old acc: 76.0, new acc: 72.5
> ```
> 
> which gave lower performance than the previous setup. Do you mind running a few sources and pasting the results here? I just like to confirm that this is an outlier.

@beijbom I think I found the problem. So `train_classifier.py` downloads the data from `spacer-trainingdata` bucket where the features were extracted by the `VGG16CaffeExtractor`. I then trained the features with `LR` and `MLP` respectively, and here is the comparison.

**LR**
```
-> Downloading data for source id: 1498.
-> Downloading 804 metadata and feature files...
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████| 804/804 [02:12<00:00,  6.09it/s]
-> Assembling train and val data for source id: 1498
-> Training...
-> Re-trained SpatSurvey (1498). Old acc: 76.0, new acc: 71.5
```

**MLP**
```
-> Downloading data for source id: 1498.
-> Downloading 804 metadata and feature files...
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████| 804/804 [00:00<00:00, 195922.64it/s]
-> Assembling train and val data for source id: 1498
-> Training...
-> Re-trained SpatSurvey (1498). Old acc: 76.0, new acc: 72.6
```

So I think it's because the current backend s1498 classifier (with 76.0% acc) was trained on 169 images at 06/22/2019 while the one you test was trained on 268 images, which means that the extra images uploaded by the user didn't help improve the performance, otherwise the training classifier status should be on the chart.

While `efficientnet_extractor.py` extracts the new features using `EfficientNetExtractor` and trains the classifier using either `LR` or `MLP`, here are the training results using `LR` and `MLP`.

**LR**
```
python scripts/regression/efficientnet_extractor.py efficientnet_b0_ver1 1498 10 LR /home/qimin/Downloads/spacer-test/test
-> Fetching 1498 image and annotation meta files...
-> Extracting features...
-> Downloading data for source id: 1498.
-> Downloading 268 metadata...
-> Assembling train and val data for source id: 1498
-> Training...
-> Re-trained SpatSurvey (1498). Old acc: 76.0, new acc: 78.9
```

**MLP**
```
python scripts/regression/efficientnet_extractor.py efficientnet_b0_ver1 1498 10 MLP /data/tmp
-> Fetching 1498 image and annotation meta files...
-> Extracting features...
-> Downloading data for source id: 1498.
-> Downloading 268 metadata...
-> Assembling train and val data for source id: 1498
-> Training...
-> Re-trained SpatSurvey (1498). Old acc: 76.0, new acc: 79.3
```
I will test on a few more sources that are not presented in the 26 test set and paste the results here.

> 
> Also: it seems we should be able to clean up the `regression` folder a bit. For example, there are two private methods for caching local, one in `efficientnet_extractor.py` and one in `train_classifier.py`. I know they do slightly different things, but can you see if it can be moved to a shared `utils.py`?

Good idea, working on it.

beijbom:
Thanks for looking into that @qiminchen . I agree this is probably b/c different number of imgs in the train data. The results on new features + MLP looks nice!

scripts/regression/utils.py
```diff
@@ -1,13 +1,42 @@
...
+from spacer import config
 from spacer.data_classes import ImageLabels
 from spacer.messages import DataLocation
 from spacer.train_classifier import trainer_factory
 
 
+def cache_local(source_root: str,
+                image_root: str,
+                export_name: str,
+                source_id: int) -> None:
```
beijbom, line 16: @qiminchen : do you want to add a boolean to say whether to cache the pre-extracted features? I don't think we need to download them for all regression tests (i.e. the new you you added)

beijbom:
@qiminchen : is this ready for final review? I just added a minor comment -- let's try to get this merged this weekend.

scripts/regression/utils.py
qiminchen: hmm what do you mean, i didnt quite get it..

qiminchen:
> @qiminchen : is this ready for final review? I just added a minor comment -- let's try to get this merged this weekend.

yes, other than the one you just added, lets get this one done and it's ready for review

scripts/regression/utils.py
beijbom: In the `train_classifier.py` regression script we need to download the features from S3 since we don't re-extract them. But we don't need the images. In `efficientnet_extractor.py` the opposite is true: we need the images but not the feature files. So I'm suggesting adding options to this caching script to just fetch what you actually need. (In all cases the metadata an labels can be fetched -- they are so small)

beijbom: Actually, looking closer at the code now I see what you did. In the `ExtractFeatures` you fetch the images directly from S3 and write to local, so you both down-load and extract at the same time. I think that is fine. But wouldn't then `utils.cache_local` over-write the extracted features?

beijbom: I think the code will be cleaner and run faster if you cache first and the do extraction and training locally.

qiminchen: > Actually, looking closer at the code now I see what you did. In the `ExtractFeatures` you fetch the images directly from S3 and write to local, so you both down-load and extract at the same time. I think that is fine.

Yes, In the `ExtractFeatures` I read the images directly from S3 and only write extracted features to local, in other words, only the extracted features will be cached to local.

> But wouldn't then `utils.cache_local` over-write the extracted features?

And no, after feature extraction, we cache the data to local, since we haven't add the boolean to specify whether to fetch feature or not, it should fetch the features from s3, but since we just extract the features (have the exact same name as what they are named in S3), features from s3 will be skipped and only metadata and labels will be fetched. So in `efficientnet_extractor.py`, `utils.cache_local` will not over-write the extracted features.
```
    for key in tqdm.tqdm(img_keys):
        _, filename = key.name.split('images')
        local_path = os.path.join(image_root, filename.lstrip('/'))
        if not os.path.exists(local_path):
            key.get_contents_to_filename(local_path)
```
In both cases, images will not be cached to local but i think its good to add a boolean to `utils.cache_local` since in `efficientnet_extractor.py`, features will not be fetched.

qiminchen: > I think the code will be cleaner and run faster if you cache first and the do extraction and training locally. This allows you to remove all s3 stuff in `ExtractFeatures` also.

so you think i should cache the images to local? because right now i didn't cache the images to local but only extracted features.

beijbom: Thanks for the explanation. I forgot there was the check to download files that already existed. Ok, so the current code should work fine then. 

You can choose - I think the code will be cleaner if you manage all AWS fetching in the `cache_local` and then operate locally from then on. And I also suspect it will run a bit faster.

qiminchen: yes, the code will be cleaner for `efficientnet_extractor.py` if we cache images, metadata and labels to local first and then operate locally from them on (extracting features and training). The only tradeoff is that the image will be downloaded and it takes up some space. I'm okay with both,  which one do you prefer? It's easy to modify the code.

beijbom: Let's do the change then. I think it will be easier to follow what is going on.

qiminchen: okay, working on it

qiminchen: okay, all AWS fetching are moved to the `cache_local`, but I didn't feel the whole procedure runs faster.
Current: download images to local + extract features
Previous: read images from S3 + extract features
I didn't test the time spend on downloading the images vs. reading images from S3, which seems to be pretty similar. Besides, downloading the images to the local takes up much space especially for large sources, each original image is in mb

beijbom:
Both
```
python train_classifier.py train 294 /data/tmp --clf_type LR
python train_classifier.py train 294 /data/tmp --clf_type MLP
```
fails with some error. @qiminchen : u mind digging around? I don't know if related to this PR, but it'd be good to understand why.

qiminchen:
> Both
> 
> ```
> python train_classifier.py train 294 /data/tmp --clf_type LR
> python train_classifier.py train 294 /data/tmp --clf_type MLP
> ```
> 
> fails with some error. @qiminchen : do you see the same error?

hmm this is weird, i did not see the error.. whats the error on your side?
Screenshot from 2020-09-20 10-04-50

beijbom:
@qiminchen : I think my error was to a feature file was only partway downloaded. I cleared the cache and it works now. I took a final pass to merge the two regression scripts. I think it's a bit cleaner now. I'm running some final tests, but this is good to merge no IMO.

beijbom:
One question: what is the recommended default number of epochs to train the MLP in your experiments?

qiminchen:
> @qiminchen : I think my error was to a feature file was only partway downloaded. I cleared the cache and it works now. I took a final pass to merge the two regression scripts. I think it's a bit cleaner now -- lmk what you think. I'm running some final tests -- once complete this is good to merge.

the cleanup looks great!

qiminchen:
> One question: what is the recommended default number of epochs to train the MLP in your experiments?

10

beijbom:
> > One question: what is the recommended default number of epochs to train the MLP in your experiments?
> 
> 10, the reason I use 10 is that the reference code of training the classifier you sent me a long time ago uses 10

Cool. That checks out.

qiminchen:
> > > One question: what is the recommended default number of epochs to train the MLP in your experiments?
> > 
> > 
> > 10, the reason I use 10 is that the reference code of training the classifier you sent me a long time ago uses 10
> 
> Cool. That checks out.

hmm weird, I can't find the code you sent me before or the link to the code, but I'm pretty sure it was 10 epoch. 5 epoch should works as well.

beijbom (approved):
Ok. Reg-tests passing nicely on my side also. Nice work!

PR #54, opened 2023-01-25
Python 3.10 support + package updates + more

StephenChan:
Software upgrades:

- Now supports Python 3.8+ (testing against 3.8 - 3.10). Dropped support for 3.6 (EOL) and 3.7 (EOL in June 2023).
- Dockerfile works at Ubuntu 22.04 (from 18.04) and Python 3.10 (from 3.6), including Caffe installation.
- Pillow >=4.2.0 to >=9.0.1 (Dockerfile on 9.4.0).
- numpy >=1.17.5 to >=1.19 (Dockerfile on 1.24.1).
- scikit-learn: ==0.22.1 to ==1.1.3. Loading models from older PySpacer versions should still work.
- torch ==1.4.0 to ==1.13.1
- torchvision ==0.5.0 to ==0.14.1
- Removed wget; replaced by `urllib.request` usage
- Removed scikit-image; wasn't being used anymore
- Updated awscli, boto3, coverage, fire, tqdm (now a dev-only requirement)

More changes:

- Started a CHANGELOG.md like coralnet (this PR bumps the version from 0.3.1 to 0.4.0)
- Previously, config variables could be specified by a `secrets.json` file or by environment variables. Now a third way is available: a `SPACER` setting in a Django project. This makes it slightly cleaner for coralnet to configure spacer (no `os.environ` calls needed in coralnet settings). Also, the `secrets.json` method no longer uses `SPACER_` prefixes for each variable name.
- The `LOCAL_MODELS_PATH` setting is now explicitly required. It was previously not required upfront, but its absence would make some tests fail (I think this was only observable if AWS access was configured and models path was not).
- PySpacer now only configures logging for fire / AWS Batch. When used as a pluggable app, it leaves the existing logging config (e.g. from coralnet) alone. This change was made by moving logging configuration statements from config.py to mailman.py.
- Removed remnants of Amazon SQS interfacing, since we always use AWS Batch now, and it's a hassle to regression-test something like SQS when we don't use it ourselves anymore.
- Added legacy-format tests for pyspacer 0.3.1 alongside the legacy-format tests for CoralNet beta, and renamed `test_beta_regression.py` to `test_legacy.py`
- Replaced setup.py with [pyproject.toml](https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/)
- Moved Amazon-ECR-related shell script(s) from this repo to the coralnet-system repo

At this point I think I understand how PySpacer connects to AWS Batch/ECR, how to update the version on PyPI, and in general how to maintain it as a package.


StephenChan:
Merged, and bumped version on PyPI to 0.4.0.

Meanwhile, I added as many git tags as I could for past releases of pyspacer, double-checking against the source distributions on PyPI. I only added a tag when there seemed to be a direct match between the corresponding PyPI release and a particular commit-history point on GitHub (only checked by inspection, nothing rigorous like a checksum).

Also deleted the 1.0 tag, since it did not correspond to any PyPI release of pyspacer. It was on PR #3 I believe (preceding the first release).

PR #58, opened 2023-08-23
README: Add new section for describing the main API + a few clarifications

StephenChan:


StephenChan:
[Minimum documentation for pyspacer's extract / train / classify calls](https://trello.com/c/bi5FSXK7/545-minimum-documentation-for-pyspacers-extract-train-classify-calls)

StephenChan:
Per issue #60, I anticipate that `TrainClassifierMsg` will get something like a `ref_labels` parameter added pretty soon, but that won't drastically change the usage.

README.md
```diff
@@ -3,37 +3,20 @@
...
 
 The spacer repo can be installed in three ways.
-* Using Docker -- the only option that supports Caffe.
+* Pip install -- for integration with other Python projects.
 * Local clone -- ideal for testing and development.
-* Using pip install -- for integration in other code-bases.
+* From Dockerfile -- the only option that supports Caffe, which is used for the legacy feature-extractor.
 
-#### Config
+### Config
```
saanobhaai, line 19: It might be good to write out somewhere in this section (perhaps at the end) the checking order, i.e. which sources trump others if multiple sources are defined.

README.md
```diff
@@ -61,10 +44,10 @@ LOCAL_MODEL_PATH is required. The two AWS access variables are required unless s
...
 To debug your configuration, try opening a Python shell and run `from spacer import config`, then `config.check()`.
 
-#### Docker build
-The docker build is the preferred build and the one used in deployment.
+### Docker build
+The docker build is used in coralnet's deployment.
 * Install docker on your system
-* Create `secrets.json` as detailed above.
+* Set up configuration as detailed above.
 * Create folder `/path/to/your/local/models` for caching model files.
```
saanobhaai, line 51: Why can't this step be done in the Dockerfile? Oh I see because the pattern is to mount a local dir? In that case maybe a pointer here would help about why you'd want to have access to cached models outside the container?

README.md
```diff
@@ -61,10 +44,10 @@ LOCAL_MODEL_PATH is required. The two AWS access variables are required unless s
...
 * Build image: `docker build -t spacer:test .`
```
saanobhaai, line 52: Why the `test` tag?

README.md
```diff
@@ -73,24 +56,221 @@ The `-v /path/to/your/local/models:/workspace/models` part will make sure
...
-#### Pip install
+### Pip install
 * `pip install pyspacer`
-* Set environmental variables.
+* Set up configuration. `secrets.json` isn't an option here, so either use environment variables, or Django settings if you have a Django project.
 
-#### Local clone
+### Local clone
 * Clone this repo
+  * If using Windows: turn Git's `autocrlf` setting off before your initial checkout. Otherwise, pickled classifiers in `spacer/tests/fixtures` will get checked out with `\r\n` newlines, and the pickle module will fail to load them, leading to test failures. However, autocrlf should be left on when adding any new non-pickle files.
```
saanobhaai, line 72: This is extremely helpful (though none of us use windows...)

README.md
```diff
@@ -73,24 +56,221 @@ The `-v /path/to/your/local/models:/workspace/models` part will make sure
...
+
+Spacer executes tasks as defined in messages. The message types are defined
+in `messages.py` and the tasks in `tasks.py`. Several data types which can be used for input and output serialization are defined
+in `data_classes.py`.
+
+For examples on how to create spacer tasks, refer to the Core API section below, and the unit tests in `test_tasks.py`.
+
+Tasks can be executed directly by calling the methods in tasks.py. 
+However, spacer also supports an interface with AWS Batch 
+handled by `env_job()` in `mailman.py`.
```
saanobhaai, line 87: Good for now, but as discussed I think decoupling should mean that no particular interface is privileged -- I should be able to invoke any of the tasks via whatever cloud system is desired. My take is, that should be via CLI and `argparse` so that the container could be run directly.

README.md
```diff
@@ -73,24 +56,221 @@ The `-v /path/to/your/local/models:/workspace/models` part will make sure
...
+
+```python
+from spacer.data_classes import ImageLabels
+from spacer.messages import DataLocation, TrainClassifierMsg
+from spacer.tasks import train_classifier
+
+message = TrainClassifierMsg(
+    # Identifier to set this training job apart from others. Should be
+    # unique for each classifier.
+    job_token='classifier1',
```
saanobhaai, line 152: Does the uniqueness apply across all jobs (and does that assume a certain `Batch` config?)? Or literally just "for each classifier"?

README.md
```diff
@@ -73,24 +56,221 @@ The `-v /path/to/your/local/models:/workspace/models` part will make sure
...
+    # 'minibatch' is currently the only trainer available.
```
saanobhaai, line 153: As opposed to what? Is this internal, or part of scikit-learn? A link would help here.

README.md
```diff
@@ -73,24 +56,221 @@ The `-v /path/to/your/local/models:/workspace/models` part will make sure
...
+    # 1. 'MLP': multi-layer perceptron; newer classifier type for CoralNet
+    # 2. 'LR': logistic regression; older classifier type for CoralNet
+    clf_type='MLP',
+    # Point-locations to ground-truth-labels (annotations) mapping. Used for
+    # training the classifier.
+    # The data dict-keys must be the same as the `key` used in the
+    # extract-features task's `feature_loc`.
+    # The data dict-values are lists of tuples of
+    # (row, column, label ID). You'll need to be tracking a mapping of
+    # integer label IDs to the labels you use.
```
saanobhaai, line 168: Good for now but hoping this can change down the road to not require ints

README.md
```diff
@@ -73,24 +56,221 @@ The `-v /path/to/your/local/models:/workspace/models` part will make sure
...
+    # Point-locations to ground-truth-labels mapping. Used for evaluating
+    # the classifier's accuracy. Should be disjoint from train_labels.
+    # CoralNet uses a 7-to-1 ratio of train_labels to val_labels.
+    val_labels=ImageLabels(data={
+        '/path/to/image4.featurevector': [(500, 2500, 1), (2500, 1500, 3)],
+        '/path/to/image5.featurevector': [(4321, 5582, 25), (4903, 2622, 19)],
+    }),
+    # Partially-filled-in DataLocation which specifies the storage type
+    # and S3 bucket (if applicable) that all the feature vectors are
+    # stored at, but doesn't specify a key.
```
saanobhaai, line 183: I'm not understanding this last point about not specifying a key

README.md
```diff
@@ -73,24 +56,221 @@ The `-v /path/to/your/local/models:/workspace/models` part will make sure
...
+    features_loc=DataLocation('filesystem', ''),
+    # List of previously-created models (classifiers) to also evaluate
+    # using this dataset, for informational purposes only.
```
saanobhaai, line 186: Interesting feature!

README.md
```diff
@@ -73,24 +56,221 @@ The `-v /path/to/your/local/models:/workspace/models` part will make sure
...
+Takes a feature vector (representing points in an image) to classify, and a classifier trained on the same type of features (EfficientNet or VGG16). Produces prediction results (scores) for the image points, as posterior probabilities for each class. Example:
+
+```python
+from spacer.messages import DataLocation, ClassifyFeaturesMsg
+from spacer.tasks import classify_features
+
+message = ClassifyFeaturesMsg(
+    # Identifier to set this classification job apart from others. Makes sense
+    # to use something that uniquely identifies the image.
+    job_token='image1',
```
saanobhaai, line 219: See note above about uniqueness

saanobhaai (changes requested):
Looking great -- just a few comments on details.

README.md
StephenChan: Good idea, will add that.

README.md
StephenChan: Now that I think of it, these pickled classifiers might be better moved to Git LFS, in which case Git might not try to do anything with newlines at all. Hopefully! (But Git LFS is for another PR.)

README.md
StephenChan: Agreed, so removing AWS Batch specifics would at least be a "finish decoupling" bullet point. CoralNet's AWS Batch workflow could be described as an example, but not tied to pyspacer on a deeper level than that.

README.md
StephenChan: Oh. I'm glad you pointed this out because yes, it's purely bookkeeping for the caller. For the CoralNet + AWS Batch workflow, the token incidentally only has to be unique across other train-classifier jobs (not across all jobs). But that shouldn't be specified in this comment as if it were a pyspacer-level requirement.

README.md
StephenChan: It's internal. Maybe Oscar considered writing a second option for a trainer class at some point, but ultimatley 'minibatch' was the only one.

README.md
StephenChan: Would this clarify better?: "This DataLocation should specify the storage_type arg, and the S3 bucket_name arg if applicable, corresponding to where the feature vectors are stored at. But the key arg isn't used."

This also implies a requirement that should get documented: all the train_labels, val_labels, and probably ref_labels later must live under the same storage_type / bucket_name.

README.md
StephenChan: I guess that doesn't have to be in these instructions. It's just that the CoralNet process for building and deploying Docker images involves tagging them.

README.md
StephenChan: There's a line below this from Oscar, which says this pattern "will make sure the downloaded models are cached to your host storage, which makes rerunning stuff much faster." Running stuff meaning building the Docker image. One such model file is 527 MB (although that's for the old Caffe extractor; the big file for the newer extractor is 66 MB).

Now, what the current Dockerfile *actually* does is redownload the models even if they already exist locally; but the local-dir mount means it *could* do an existence check and avoid redownloading. Also, it does still save on storage space, by not having to store one copy of the models per Docker image (when developing/testing on Docker, it'll be common to have multiple built images stored at a given time).

But this isn't necessary for a production build to work. Even in CoralNet's use cases, the outside-container environment would only be available when testing the Docker image in the environment where it was built, not when the Docker image is running on instances spun up by AWS Batch (through mailman.py).

So perhaps the models-local-dir mount should be an optional step in pyspacer's Dockerfile(s).

The second local-dir mount may also not be necessary if using environment variables instead of secrets.json. Certainly some things to think about, but all this Dockerfile splitting/reworking would be for a separate PR.

README.md
saanobhaai: Yes that sounds good -- maybe `The key arg isn't used, because the location is for all feature vectors, not any particular key."

README.md
StephenChan: See commit 41c6724

README.md
StephenChan: See commit 6682a39 - generalized `spacer:test` to `myimagename`, and tried to clarify some surrounding instructions too.

README.md
StephenChan: See commit be57ec7

README.md
StephenChan: See commit 582c17d

README.md
StephenChan: See commit 83b70f6, trying different wording: "All the feature vectors should use the same storage_type, and the same S3 bucket_name if applicable. This DataLocation's purpose is to describe those common storage details. The key arg is ignored, because that will be different for each feature vector."

saanobhaai (approved):
This is great Stephen, thank you. Let's merge and move on!

PR #62, opened 2023-09-27
Allow extractors to be loaded from anywhere, not just CoralNet S3

StephenChan:
Currently you need CoralNet S3 access in order to extract features at all. So, here we try to fix that.

Update:

Generalize feature extractor support by allowing any FeatureExtractor subclass instance, and any file locations instead of just from CoralNet's S3 bucket (which requires auth).

- Test extractors/fixtures are still hidden away in CoralNet infrastructure for now. That's something to address in the future.
- Filesystem-caching now applies to any S3/URL extractor-file locations.
- SHA256-checking for extractor files is still supported, as an optional `data_hashes` in the `FeatureExtractor` constructor.
- Rework relevant config vars. Of note, spacer is now usable without any specified config, if neither S3 or remote-loaded extractors are needed.
- Remove awscli dependency in Dockerfile, because 1) the Dockerfile commands that required awscli no longer apply, and 2) there's currently a Cython 3.x related problem with installing PyYAML, a dependency of awscli (latest PyYAML, 6.0.1, tries to work around it by pinning Cython below 3, but even better to not worry about it at all).

Code/design notes off the top of my head:

- `extract_features.py` has a fair bit of code related to storage now. Maybe some of this should be factored out into `storage.py` later.
- Maybe there should end up being a `File` or `Blob` class which is more general than `DataLocation`. spacer's 'pure' concept of `DataLocation` is to be a location of a `DataClass`. `FeatureExtractor` isn't a `DataClass` as it's not just a vessel for data, it also has a method to extract features. However, `FeatureExtractor` isn't the only instance of the code where the `DataLocation` concept is stretched, with the code reaching directly into the `DataLocation` fields rather than calling its methods. So making a cleaner design would take a few steps. In the interest of time, I put it off for the moment.

### DIFF
--- a/spacer/messages.py
+++ b/spacer/messages.py
@@ -4,7 +4,10 @@ Each data-class can serialize itself to a structure of JSON-friendly
 python-native data-structures such that it can be stored.
 """
 
+from __future__ import annotations
+from pathlib import Path
 from typing import List, Tuple, Dict, Union, Optional
+from urllib.parse import urlparse
 
 from spacer import config
 from spacer.data_classes import DataClass, ImageLabels
@@ -32,6 +35,23 @@ class DataLocation(DataClass):
     def example(cls) -> 'DataLocation':
         return DataLocation('memory', 'my_blob')
 
+    @property
+    def filename(self) -> str:
+        if self.storage_type == 'url':
+            # This is a basic implementation which just gets the last
+            # part of the URL 'path', even if that part isn't filename-like.
+            return Path(urlparse(self.key).path).name
+        else:
+            return Path(self.key).name
+
+    @property
+    def is_remote(self) -> bool:
+        return self.storage_type in ['url', 's3']
+
+    @property
+    def is_writable(self) -> bool:
+        return self.storage_type != 'url'
+
     @classmethod
     def deserialize(cls, data: Dict) -> 'DataLocation':
         return DataLocation(**data)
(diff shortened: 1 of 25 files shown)


spacer/messages.py
```diff
@@ -41,74 +61,79 @@ def __hash__(self):
...
         self.job_token = job_token
-        self.feature_extractor_name = feature_extractor_name
+        self.extractor = extractor
         self.rowcols = rowcols
         self.image_loc = image_loc
         self.feature_loc = feature_loc
 
+    @classmethod
+    def example(cls) -> 'ExtractFeaturesMsg':
+        from spacer.extract_features import EfficientNetExtractor
```
saanobhaai, line 89: I'm probably just missing something but why not do the imports in the usual place up top?

saanobhaai (commented):
Just a small comment; otherwise looks like huge improvements to me.

spacer/messages.py
StephenChan: `extract_features.py` imports `messages`, so doing this method-level import avoids an unresolvable circular import situation.

StephenChan:
[Allow extractors to be loaded from anywhere, not just CoralNet S3](https://trello.com/c/2KiSPNg4/546-allow-extractors-to-be-loaded-from-anywhere-not-just-coralnet-s3)

Issue #60, opened 2023-09-17
Training reference set is inflexible, making label inclusion unpredictable


StephenChan:
As seen in train_utils.py's `train()` function ([link](https://github.com/coralnet/pyspacer/blob/0.4.1/spacer/train_utils.py#L20)):

- The `labels` (which were the `train_labels` of the TrainClassifierMsg) are split up into a train set and a reference set. The reference set is determined by taking every 10th image of `labels`, up to 5000 annotations (assuming the number of samples per image is constant throughout `labels`). The 5000 cap is to ensure the reference set fits in memory. The reason the reference set is made to be separate from the train set is that it's used to calibrate the classifier (according to the code; I can't provide a deeper explanation myself).

- The classifier's set of recognized classes (i.e. its labelset) ends up being the set of classes present in both the train set and reference set.

As a result, from the point of view of the program calling pyspacer, it's difficult to control what the classifier's recognized labelset ends up being.

Also, if the rarity of a class is on the order of 1 in 5000 points, there's a good chance that class ends up being excluded from the classifier's recognized labelset. Further, if the labelset size is on the order of 5000 labels, most of the labels have a good chance of being excluded. That's a significant obstacle to any vision of, say, a generalized CoralNet-wide classifier (CoralNet has 7000+ labels).

Possible solution avenues; more than one might be desired:

1. Make the number 5000 customizable, so that a system with higher-memory instances than CoralNet can raise to potentially a much higher number.

2. Rework training so that the reference set doesn't have to fit in memory and thus doesn't have to be size-capped at all. Not sure how doable this is. If there are any caveats to this training implementation compared to the old one, then perhaps the new implementation would become a new Trainer class (alongside the existing MiniBatchTrainer).

3. Allow TrainClassifierMsg to specify train labels, val labels, *and* ref labels - not just train and val. This means the reference set is much more clearly defineable (don't have to set it up as every 10th element of your train labels). To reduce entry-level complexity, maybe this granularity can be optional: have the option to specify train & val & ref, and also the option to just specify 'labels' and let pyspacer figure out how to split it up into train/val/ref. I think this solution should be implemented, because CoralNet still has trouble handling the corner case where the classifier's recognized labelset is size 0, and being able to reliably detect that situation before calling pyspacer (by defining the ref set on CoralNet's side) would help a lot.

StephenChan:
[...]

Issue #59, opened 2023-09-17
Training mini-batch size is bespoke to CoralNet instance specs


StephenChan:
As seen in train_utils.py's `train()` function ([link](https://github.com/coralnet/pyspacer/blob/0.4.1/spacer/train_utils.py#L28)), the mini-batch size (and reference-set size cap) is determined as follows:

```python
# Calculate max nbr images to keep in memory (for 5000 samples total).
max_imgs_in_memory = 5000 // labels.samples_per_image
```

Better to base this on a customizable number instead of a hardcoded 5000.

StephenChan:
Done in PR #71. Now controlled by the config var `TRAINING_BATCH_LABEL_COUNT`.

PR #71, opened 2023-12-06
Training: specifying ref set + filtering by class + better batching

StephenChan:
For issue #60.

- Training labels are now specified either as just one set, or as train + reference + val sets.
- Training mini-batch size is now configurable as `TRAINING_BATCH_LABEL_COUNT`.
- Training mini-batch size is now followed more accurately, rather than approximating based on number of labels in the first image.

Note: I'll probably end up amending commit 813a17b to update tests / fix any bugs.

### DIFF
--- a/spacer/task_utils.py
+++ b/spacer/task_utils.py
@@ -109,7 +109,7 @@ def preprocess_labels(
         ref_done = False
 
         for image_index, image_key in enumerate(labels_in.image_keys):
-            this_image_labels = labels_in.data[image_key]
+            this_image_labels = labels_in[image_key]
 
             if image_index % 10 == 0:
                 # 1st, 11th, 21st, etc. images go in val.
--- a/spacer/task_utils.py
+++ b/spacer/task_utils.py
@@ -1,9 +1,15 @@
 from __future__ import annotations
+from logging import getLogger
 
 from PIL import Image
 
 from spacer import config
-from spacer.exceptions import DataLimitError, RowColumnInvalidError
+from spacer.data_classes import ImageLabels
+from spacer.exceptions import (
+    DataLimitError, RowColumnInvalidError, TrainingLabelsError)
+from spacer.messages import TrainingTaskLabels
+
+logger = getLogger(__name__)
 
 
 def check_extract_inputs(image: Image,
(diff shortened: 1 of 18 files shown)


spacer/task_utils.py
```diff
@@ -109,7 +109,7 @@ def preprocess_labels(
         ref_done = False
 
         for image_index, image_key in enumerate(labels_in.image_keys):
-            this_image_labels = labels_in.data[image_key]
+            this_image_labels = labels_in[image_key]
```
saanobhaai, line 112: Here an elsewhere, doing `[image_key]` rather than `.get(image_key)` means a KeyError exception if for some reason the key doesn't exist -- that's probably what you want, but just checking; you could handle or assign a default if that would help debugging.

StephenChan: Good to check - I think in each of these cases it's getting `image_key` by iterating over the `image_keys` property of the same `ImageLabels` object, so it ought to be safe enough IMO; I also think there isn't really a reasonable default value to use, so it may as well raise a KeyError if the key's not there.

StephenChan:
[pyspacer training: Make reference set larger & specifiable to ensure rare labels are recognized](https://trello.com/c/EWhZrA9U/560-pyspacer-training-make-reference-set-larger-specifiable-to-ensure-rare-labels-are-recognized)

StephenChan:
[code to filter annotations by desired set of labels](https://trello.com/c/FTtdhqzd/629-code-to-filter-annotations-by-desired-set-of-labels)

StephenChan:
So there were a few issues pertaining to training which all involved code/logic related to each other, and I ended up addressing them at the same time in this PR. Those issues would be #60, #59, #78, and a feature to filter annotations by a desired set of labels (indicated by the Trello link above, and self-explanatory from the title I think).

The main changes since this issue's OP involve how to build the `labels` argument of `TrainClassifierMsg`. I've pasted the relevant part of the updated README here:

> The labels must be split into training, reference, and validation sets:
> 
> - The training set (train) is the data that actually goes into the classifier training algorithm during each training epoch. This is generally much larger than the other two sets.
> - The reference set (ref) is used to evaluate and calibrate the classifier between epochs. 
> - The validation set (val) is used to evaluate the final classifier after training is finished.
> 
> This three-set split is known by other names elsewhere, such as [training, validation, and test sets](https://en.wikipedia.org/wiki/Training%2C_validation%2C_and_test_data_sets) respectively, or [training, development, and test sets](https://cs230.stanford.edu/blog/split/) respectively.
> 
> There are a few ways to create the `labels` structure. Each way involves creating one or more instances of `data_classes.ImageLabels`:
> 
> ```python
> from spacer.data_classes import ImageLabels
> image_labels = ImageLabels({
>     # Labels for one feature vector's points.
>     '/path/to/image1.featurevector': [
>         # Point location at row 1000, column 2000, labeled as class 1.
>         (1000, 2000, 1), 
>         # Point location at row 3000, column 2000, labeled as class 2.
>         (3000, 2000, 2),
>     ],
>     # Labels for another feature vector's points.
>     '/path/to/image2.featurevector': [
>         (1500, 2500, 3),
>         (2500, 500, 1),
>     ],
> })
> ```
> 
> The `labels` argument of `TrainClassifierMsg` expects an instance of `data_classes.TrainingTaskLabels`. There are a few ways to create this:
> 
> 1. Pass a single ImageLabels instance to the `task_utils.preprocess_labels()` function. preprocess_labels() then decides how to split up your labels into train, ref, and val (while doing error checks in the meantime), and creates a TrainingTaskLabels instance from there.
> 2. Pass three ImageLabels instances to the TrainingTaskLabels constructor: one instance for each of train, ref, and val.
> 3. Do method 1, but also specify the `accepted_classes` argument to preprocess_labels(); this makes the function filter out any labels that aren't in the desired set of classes.
> 4. Do method 2, but also pass the TrainingTaskLabels through preprocess_labels(). This allows you to use the error-checking and accepted_classes parts of preprocess_labels(), and the train/ref/val split you defined will remain intact.
> 
> ```python
> from spacer.data_classes import ImageLabels
> from spacer.messages import TrainingTaskLabels
> from spacer.task_utils import preprocess_labels
> 
> # 1
> labels = preprocess_labels(ImageLabels(...))
> # 2
> labels = TrainingTaskLabels(
>     train=ImageLabels(...), ref=ImageLabels(...), val=ImageLabels(...))
> # 3
> labels = preprocess_labels(ImageLabels(...), accepted_classes={...})
> # 4
> labels = preprocess_labels(TrainingTaskLabels(...), accepted_classes={...})
> ```

Other comments after working on this decently involved PR:

- I was wondering if an overall move in terminology from 'labels' to 'annotations' (to mean 'the application of a label/class to a point') would make things less confusing. Potentially also 'classes' to 'labels' to match CoralNet, but not sure. The terminology shift would involve a lot of little changes throughout spacer, so it should be its own PR if we do it.
- Maybe some function names could be more clear, like `load_train_data_for_image()` instead of `load_image_data()`.
- I was tempted to simplify out `trainer_name` and `trainer_factory` because there's still only one available trainer name, but I ended up saving it for a future PR. Relatedly, some of the task messages' arguments could be optional or have reasonable defaults. For example, `previous_model_locs` could be optional, and `clf_type` could default to `'MLP'`.
- I'm increasingly wishing for a way to override settings (config) values for certain unit tests, the way Django has it, so that warrants an issue at some point. Right now, the tests which depend on particular settings simply assume that the setting has been left as the default value in the current environment, which makes test runs potentially a hassle for anyone using non-default values. Also, being able to set certain settings to a bare minimum value can speed up tests; for example, a test which depends on TRAINING_BATCH_LABEL_COUNT would likely run faster if the setting was 50 instead of 5000.
- The `scripts` dir is getting increasingly out of date, as we don't have tests for them and I still haven't used them in a while... so that needs cleaning up one way or the other sometime.

Issue #73, opened 2023-12-07
Train time seems to depend heavily on image count


StephenChan:
Some recent train times:

- 1052 (VGG16) completed in 52 hours: 56,000 images x 30 points
- 1656 (VGG16) completed in 50 hours: 65,000 images x 50 points
- 295 (VGG16) completed in 38 hours: 63,000 images x 10 points
- 1374 (VGG16) completed in 21 hours: 34,000 images x 30 points
- 1288 (EffNet) completed in 7 hours: 29,000 images x 10 points
- 3182 (EffNet) completed in 1.6 hours: 3800 images x 100 points
- 3342 (EffNet) completed in 1.5 hours: 1200 images x 1000 points

There are a few more train times in issue #53 too.

Looking at 1052 vs. 1656, obviously there are just some fluctuations on how fast training jobs run, but overall the points per image doesn't seem like as big of a factor as one might expect. Especially looking at the EfficientNet times which we should be more interested in going forward. It's definitely got some effect, but not close to linear. On the other hand, the image count has a closer to linear influence than we might expect, given that the number of training batches is determined by the number of points (features), and given that there's no parallelization in the whole training process.

And there are no anomalies with the feature vector filesize; that's pretty much linearly related to point count, at least for EfficientNet:

- EffNet 10 points (1288) examples: i3653915 = 56 KB, i1108305 = 56 KB
- EffNet 100 points (3182) examples: i3655456 = 546 KB, i3655386 = 546 KB
- EffNet 1000 points (3342) examples: i3279111 = 5.3 MB, i3278833 = 5.3 MB

As for what timescale we're talking about per image:

- 1052: 52 hours / 56,000 images = 3.3 seconds/image
- 1656: 50 hours / 65,000 images = 2.8 seconds/image
- 295: 38 hours / 63,000 images = 2.2 seconds/image
- 1374: 21 hours / 34,000 images = 2.2 seconds/image
- 1288: 7 hours / 29,000 images = 0.9 seconds/image
- 3182: 1.6 hours / 3800 images = 1.5 seconds/image
- 3342: 1.5 hours / 1200 images = 4.5 seconds/image

Keep in mind that we always run 10 training epochs, so the seconds per image per epoch would be 0.33, 0.28, etc.

Here's the first thing I'm thinking of looking into, from issue #53:

> Try to reuse S3 connections as long as possible instead of using one connection per feature vector download

[...]

PR #77, opened 2023-12-12
Reuse S3 connection (resource) through thread's lifetime

StephenChan:
Here's the first thing I wanted to look into for issue #73.

`boto3.resource('s3', <credentials>)` was being called once for each feature-vector download (once per image per epoch) during training. Training was taking somewhere between 0.1-0.3 seconds per image per epoch in many cases, and a connection/authentication step like this seemed like a potentially significant factor at that timescale.

I wanted to see if that resource could be reused between feature-vector downloads, or indeed between any uses of boto's S3 API within the same thread. So I tested it, and yes, we can reuse the resource. I don't know if it expires after a certain time, but so far it's held up through a training of 60,000 images x 2 epochs (120,000 uses), and (in a separate test) about 18 hours. I'm still testing the longevity aspect to see if the resource lasts at least a day or two.

As for the training performance improvement, see the following CSV of experiments (all EfficientNet + MLP). "reuse conn" is the new code, "new conn" is the old code:

[train-time_s3-connection-reuse_2023-12.csv](https://github.com/coralnet/pyspacer/files/13642124/train-time_s3-connection-reuse_2023-12.csv)

(Abridged data, since the whole thing doesn't fit well in a GitHub thread:)

| Source | Images | Points/im | Epochs | Time/im/epoch: reuse conn | Time/im/epoch: new conn | Time-diff/im/epoch | Speedup ratio |
|--------|--------|-----------|--------|---------------------------|-------------------------|--------------------|---------------|
| 1097   | 1686   | 10        | 10     | 0.046                     | 0.113                   | 0.067              | 2.48          |
| 1097   | 1686   | 10        | 2      | 0.061                     | 0.133                   | 0.072              | 2.182         |
| 1097   | 1686   | 10        | 2      | 0.055                     | 0.128                   | 0.073              | 2.312         |
| 1388   | 8706   | ~8.5      | 2      | 0.084                     | 0.158                   | 0.075              | 1.889         |
| 3064   | 2211   | 25        | 2      | 0.094                     | 0.182                   | 0.088              | 1.928         |
| 3413   | 5079   | 25        | 2      | 0.109                     | 0.179                   | 0.07               | 1.638         |
| 2138   | 720    | 49        | 2      | 0.087                     | 0.182                   | 0.095              | 2.1           |
| 3354   | 1600   | 50        | 2      | 0.106                     | 0.195                   | 0.088              | 1.832         |
| 1933   | 560    | 100       | 2      | 0.112                     | 0.189                   | 0.077              | 1.691         |
| 3520   | 253    | 200       | 2      | 0.115                     | 0.189                   | 0.074              | 1.639         |
| 1360   | 49     | 1000      | 2      | 0.313                     | 0.358                   | 0.045              | 1.143         |

So with coralnet's current infrastructure, the change saves roughly 0.07 seconds per image per epoch, and overall speeds up training by 1.6-2.2x. There's some variance of unknown origin, maybe just AWS server load, as seen in the multiple tests on source 1097. But I'd say the speedup is still fairly clear.

The time per image per epoch still isn't very close to being linearly related to the points per image, so there's still more to be looked at with issue #73, but this is a start.

### DIFF
--- a/spacer/config.py
+++ b/spacer/config.py
@@ -150,20 +150,39 @@ if LOG_DESTINATION:
     )
 
 
+S3_CONNECTION = None
+
+
 def get_s3_conn():
     """
     Returns a boto s3 connection.
-    - It first looks for credentials in spacer config.
-    - If not found there it will default to credentials in ~/.aws/credentials
+    Each thread only establishes a connection once, saving it to the global
+    S3_CONNECTION and reusing it thereafter.
     """
     if not all([AWS_REGION, AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY]):
         raise ConfigError(
             "All AWS config variables must be specified to use S3.")
 
-    return boto3.resource('s3',
-                          region_name=AWS_REGION,
-                          aws_access_key_id=AWS_ACCESS_KEY_ID,
-                          aws_secret_access_key=AWS_SECRET_ACCESS_KEY)
+    global S3_CONNECTION
+
+    if S3_CONNECTION is None:
+
+        # This passes credentials from spacer config. If credentials are
+        # None, it will default to using credentials in ~/.aws/credentials
+        S3_CONNECTION = boto3.resource(
+            's3',
+            region_name=AWS_REGION,
+            aws_access_key_id=AWS_ACCESS_KEY_ID,
+            aws_secret_access_key=AWS_SECRET_ACCESS_KEY,
+        )
+
+        # We're interested in confirming:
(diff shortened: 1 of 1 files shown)


StephenChan:
Reworked so that the connection's not shared between threads (that was the original goal, but I did it wrong) and doesn't need the `global` keyword.

StephenChan:
OK, with the latest code, the same connection/resource was still able to be reused over 2 days later. Let's merge it.

PR #80, opened 2023-12-20
Training: cache feature vectors for 2nd epoch onward

StephenChan:
Cache feature vectors in the local filesystem if they were loaded from remote storage (S3 or URL).

I *think* this will do what it intends to, but I couldn't figure out a way to do it without committing OOP crimes somewhere (`data_classes.ImageFeatures.load()` in this case). I think the better way to do this would involve passing a `Storage` to TrainClassifierMsg instead of a `DataLocation`, and putting the temporary directory attribute onto that Storage. That probably involves a much larger refactor overall where the designs of `Storage`, `DataLocation`, and perhaps `DataClass` are reworked. I have ideas for that but I don't think I'm up for that for the remainder of the month.

Also want to make the caching optional (but on by default), so that it can be turned off in case filesystem space is a concern, as it may be for coralnet's largest sources (particularly since older sources have feature vectors about 8x in filesize).

Let me know if it'd be useful to merge this PR in the short term though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 - The `train_classifier` task now accepts label IDs as either integers or strings, not just integers.
 
+- The `train_classifier` task is now able to locally cache feature vectors which were loaded from remote storage, which can greatly speed up training from epoch 2 onward. This is optional and enabled by default; the location of the cache directory is also configurable.
+
 ## 0.8.0
 
 - `ImageFeatures` with `valid_rowcol=False` are no longer supported for training. For now they are still supported for classification.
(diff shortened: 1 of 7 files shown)


StephenChan:
| Source | Images | Points/im | Epochs | Time for epoch 1 | Time for each subsequent epoch
|--------|--------|-----------|--------|---------------------------|-------------------------|
| 1097   | 1686   | 10        | 3     | 96s                     | 3s                   |
| 3354   | 1600   | 50        | 10     | 178s                     | 8s                   |

The code may be ugly so far, but the results sure aren't... (remember, each subsequent epoch used to be as long as epoch 1)

StephenChan:
[Feature caching PR](https://trello.com/c/Uzt0BF88/703-feature-caching-pr)

StephenChan:
Ready for review. Recent changes:

- Rebased on top of recently merged PRs
- Merged the two old commits (...since the second commit message had a typo which was bugging me)
- Added options for the caching behavior: OS-designated temp dir location, custom temp dir location, or no caching. No caching is useful if the caching could cause disk-space problems.
- Unit test improvement
- Updated README and CHANGELOG

Still want to refactor adjacent classes/methods for better OOP design, but not urgent and will leave that for another time.

yeelauren (approved):

PR #85, opened 2024-01-22
Improved training sampling + cache features (combining two WIP branches)

StephenChan:
Combo of PR #84 and PR #80 in their current WIP states.

PR #88, opened 2024-01-24
Fix bucket_name bucketname consistency

yeelauren:
Separating out into smaller PRs  :smile_cat: 

Refactor for consistency see: https://github.com/coralnet/pyspacer/blob/cb0b427b03f60bb92e2e9bdf1ccbb64e5b2230c3/spacer/storage.py#L187

https://github.com/coralnet/pyspacer/blob/cb0b427b03f60bb92e2e9bdf1ccbb64e5b2230c3/spacer/storage.py#L205

https://github.com/coralnet/pyspacer/blob/cb0b427b03f60bb92e2e9bdf1ccbb64e5b2230c3/spacer/messages.py#L25

### DIFF
--- a/spacer/storage.py
+++ b/spacer/storage.py
@@ -98,8 +98,8 @@ class URLStorage(Storage):
 class S3Storage(Storage):
     """ Stores objects on AWS S3 """
 
-    def __init__(self, bucketname: str):
-        self.bucketname = bucketname
+    def __init__(self, bucket_name: str):
+        self.bucket_name = bucket_name
         # Prevent `RuntimeError: cannot schedule new futures after
         # interpreter shutdown`.
         # Based on https://github.com/etianen/django-s3-storage/pull/136
@@ -107,23 +107,23 @@ class S3Storage(Storage):
 
     def store(self, key: str, stream: BytesIO):
         s3 = config.get_s3_conn()
-        s3.Bucket(self.bucketname).put_object(Body=stream, Key=key)
+        s3.Bucket(self.bucket_name).put_object(Body=stream, Key=key)
 
     def load(self, key: str):
         s3 = config.get_s3_conn()
         stream = BytesIO()
-        s3.Object(self.bucketname, key).download_fileobj(
+        s3.Object(self.bucket_name, key).download_fileobj(
             stream, Config=self.transfer_config)
         return stream
 
     def delete(self, key: str) -> None:
         s3 = config.get_s3_conn()
-        s3.Object(self.bucketname, key).delete()
+        s3.Object(self.bucket_name, key).delete()
 
     def exists(self, key: str):
         s3 = config.get_s3_conn()
         try:
-            s3.Object(self.bucketname, key).load()
+            s3.Object(self.bucket_name, key).load()
             return True
         except botocore.exceptions.ClientError:
             return False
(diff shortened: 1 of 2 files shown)


yeelauren:
Additionally a new change from #85 ircc will also need to be changed  https://github.com/coralnet/pyspacer/blob/478a8f6f01d7bdc1cb7b40866d36d89b53dda314/spacer/messages.py#L71

StephenChan:
Thanks Lauren! This'll make reviewing easier.

I'm definitely onboard with making this param name consistent, but:

1. It looks like there are still a decent amount of `bucket_name` usages remaining: for example in `messages.py`, `storage.py`, `tests/test_storage.py`, `tests/utils.py`, and `README.md`.
2. Pardon my pickiness, but unless there's another reason I've overlooked (e.g. consistency with boto), I think I'd actually favor `bucket_name` over `bucketname`. The style goes better with the `storage_type` param in `DataLocation.__init__()` and `storage_factory()`; and originally there were more usages of `bucket_name` than `bucketname` (around 50 vs. 10).

yeelauren:
@StephenChan I agree that makes more sense - the latest push should address those changes !

spacer/tasks.py
```diff
@@ -3,8 +3,7 @@
 """
 import time
 import traceback
-from logging import getLogger
-
```
StephenChan, line 7: This `getLogger` import is still needed. (This was something I worked with recently, so maybe this deletion was a merge/rebase slip-up?)

spacer/storage.py
```diff
@@ -184,12 +184,14 @@ def clear_memory_storage():
...
 
-def storage_factory(storage_type: str, bucketname: str | None = None):
+def storage_factory(storage_type: str, bucket_name: str | None = None):
 
     assert storage_type in config.STORAGE_TYPES
 
     if storage_type == 's3':
-        return S3Storage(bucketname=bucketname)
+        if bucket_name is None:
+            raise ValueError("bucket_name must be a string for s3 storage")
```
StephenChan, line 193: This check works for me since it's related and useful.

spacer/tasks.py
```diff
@@ -28,8 +27,14 @@ def extract_features(msg: ExtractFeaturesMsg) -> ExtractFeaturesReturnMsg:
 
     check_extract_inputs(img, msg.rowcols, msg.image_loc.key)
 
+    check_rowcols(msg.rowcols, img)
```
StephenChan, line 30: `check_rowcols()` was replaced with `check_extract_inputs()` in PR #70. So don't need this line.

spacer/tasks.py
```diff
@@ -28,8 +27,14 @@ def extract_features(msg: ExtractFeaturesMsg) -> ExtractFeaturesReturnMsg:
...
+    if not isinstance(msg.extractor, Callable):
+        raise TypeError("msg.extractor must be callable")
+    
     with config.log_entry_and_exit('actual extraction'):
         features, return_msg = msg.extractor(img, msg.rowcols)
+    if not hasattr(features, 'store'):
+        raise TypeError("features object must have a store method")
```
StephenChan, line 37: These two type checks (well, instance/attribute checks to be exact) aren't that related to the rest of the PR.

They're also not really needed in my opinion - I think in Python, type checks don't hurt individually but they can add up to clutter the codebase somewhat. I know Oscar had several such checks in the codebase here and there, but in general I wasn't really keen on adding more. Sometimes they're a good idea with recent API changes (like PR #62 which admittedly relates to the first check here), but I'd like to think that's more of a worry for apps with larger userbases than pyspacer's.

StephenChan (changes requested):
Thanks Lauren, looks good regarding the param naming! I reviewed the other bits of the PR.

StephenChan (approved):
Thanks!

PR #89, opened 2024-01-24
Update torch_utils.py to use gpu if available

yeelauren:


StephenChan:
I'm fine with this. In the future, might be good to make this optional, in case GPU is available but still not desired in a particular environment - for reasons of resource balancing, heat/fan noise, cost, etc. But coralnet shouldn't care at this point since its Batch-spun EC2 instances don't have GPUs.

(Just note that pyspacer currently only uses PyTorch for feature extraction, not training or classification.)

I'll go ahead and merge this if that sounds good.

PR #94, opened 2024-02-29
(See PR #95 instead)

StephenChan:
Edit: Replaced with PR #95


StephenChan:
[pyspacer: accept any hashable data type (unique) as label id, not just integer](https://trello.com/c/JaCtYm6Z/556-pyspacer-accept-any-hashable-data-type-unique-as-label-id-not-just-integer)

PR #96, opened 2024-03-07
Supported Pythons: 3.8 to 3.10 -> 3.10 to 3.11

StephenChan:
Doing this because I keep accidentally relying on Python 3.10 stuff (mainly related to type annotations), and there currently doesn't seem to be a need for 3.8/3.9 support (CoralNet and MERMAID use 3.10). And 3.11 support didn't seem like that much of a stretch.

torch and torchvision accepted versions have been relaxed to accommodate Python 3.11 (torch==1.13.1 to torch>=1.13.1,<2.3; torchvision==0.14.1 to torchvision>=0.14.1,<0.18). Hopefully that works out in GitHub's CI.

If CI passes, I'll just merge this whenever my next PR is ready to go; but in the meantime, reviews are welcome.

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -9,10 +9,11 @@ on:
 jobs:
   build:
 
-    runs-on: ubuntu-22.04
+    # 23.04 isn't an LTS version, but ensures that Python 3.11 is available
+    runs-on: ubuntu-23.04
     strategy:
       matrix:
-        python-version: [ "3.8", "3.9", "3.10" ]
+        python-version: [ "3.10", "3.11" ]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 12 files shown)


StephenChan:
I did end up making some changes to the CI structure. Instead of using a single `requirements.txt` which is tested on each supported Python version, now there are 3 requirements files:

- `dev.txt`, package versions recommended for developers; this is all we had previously
- `min.txt`, pyproject.toml's minimum accepted versions
- `max.txt`, pyproject.toml's maximum accepted versions

And now CI runs those on Python 3.10, 3.10, and 3.11 respectively. The advantage over the old way is that now the pyproject.toml version-ranges are CI-tested. I'm not sure if this way of doing it reflects a common practice in Python package development, but it makes sense to me at the moment, at least.

StephenChan:
As stated previously:

> If CI passes, I'll just merge this whenever my next PR is ready to go

Issue #74, opened 2023-12-07
Training with GPU; requires replacing scikit-learn


StephenChan:
There's very much a question of whether this is worth the leap for us in the short term, but either way it's worth documenting what this might involve.

First of all, we use scikit-learn for running training, whether it's MLP (multilayer perceptron) or LR (logistic regression) / SGD (stochastic gradient descent) training.

Second, scikit-learn is CPU-only and their MLP is apparently on the simple side:

- [scikit-learn FAQ](https://scikit-learn.org/stable/faq.html#will-you-add-gpu-support):

  > Will you add GPU support?
  > No, or at least not in the near future. The main reason is that GPU support will introduce many software dependencies and introduce platform specific issues. scikit-learn is designed to be easy to install on a wide variety of platforms. Outside of neural networks, GPUs don’t play a large role in machine learning today, and much larger gains in speed can often be achieved by a careful choice of algorithms.

  Note that MLP = neural network.

  > Deep learning and reinforcement learning both require a rich vocabulary to define an architecture, with deep learning additionally requiring GPUs for efficient computing. However, neither of these fit within the design constraints of scikit-learn; as a result, deep learning and reinforcement learning are currently out of scope for what scikit-learn seeks to achieve.
  >
  >You can find more information about addition of gpu support at [Will you add GPU support?](https://scikit-learn.org/stable/faq.html#will-you-add-gpu-support).
  >
  > Note that scikit-learn currently implements a simple multilayer perceptron in [sklearn.neural_network](https://scikit-learn.org/stable/modules/classes.html#module-sklearn.neural_network). We will only accept bug fixes for this module. If you want to implement more complex deep learning models, please turn to popular deep learning frameworks such as [tensorflow](https://www.tensorflow.org/), [keras](https://keras.io/) and [pytorch](https://pytorch.org/)

- [scikit-learn neural networks page](https://scikit-learn.org/stable/modules/neural_networks_supervised.html):

  > Warning
  > This implementation is not intended for large-scale applications. In particular, scikit-learn offers no GPU support. For much faster, GPU-based implementations, as well as frameworks offering much more flexibility to build deep learning architectures, see Related Projects.

- [A Stack Overflow Q&A](https://stackoverflow.com/questions/54527439/differences-in-scikit-learn-keras-or-pytorch):

[...]

PR #97, opened 2024-03-10
Enhance annotation sampling for training, v2

StephenChan:
Newer version of PR #84. The difference is that this PR has gone on top of the merges of 95 and 96, resolving any conflicts. This is a new PR because I wanted to leave the old `training-annotation-sampling` branch intact, in case it's still being used for some tests right this moment.

This PR is ready for review, and is the next thing I'm looking to finally merge.

Per the updated CHANGELOG:

> `task_utils.preprocess_labels()` now has three available modes on how to split training annotations between train, ref, and val sets. Differences between the three modes - `VECTORS`, `POINTS`, and `POINTS_STRATIFIED` - are explained in the `SplitMode` Enum's comments. Additionally, all three modes now ensure that the ordering of the given training data has no effect on which data goes into train, ref, and val.

And here are said Enum's comments:

```python
class SplitMode(Enum):
    """
    How to split annotations between train, ref, and val sets.
    """
    # Each feature vector's points all go into train, or all go into ref, or
    # all go into val.
    #
    # The rationale behind this mode is to have greater separation between
    # training data and evaluation data, whether it's during the calibration
    # process (train vs. ref) or during evaluation of the final classifier
    # (train vs. val).
    # Here we assume the imagery is 'more different' when going across feature
    # vectors, as opposed to staying within the same feature vector. When
    # training and evaluation data are 'more different', the result is more
    # useful.
    # Thus, this mode can improve usefulness of calibration, and rigor of the
    # evaluation results.
    # However, the annotation count may not end up precisely balanced
    # between train/ref/val as desired, particularly when the feature vector
    # size is comparable to the set size. For example, if each feature vector
    # has 100 points, and the target ref-set size is 450, then the best we can
    # do is giving the ref set either 400 or 500 points.
    VECTORS = 'vectors'
    # The split is done on an individual point basis, so a single
    # feature vector may be split across train/ref/val.
    #
    # This allows the annotation count to be more precisely balanced
    # between train/ref/val.
    # However, there may be concerns that the imagery going into each set is
    # too similar, particularly when points are densely distributed within
    # each image.
    POINTS = 'points'
    # Stratified sampling by class: an A%/B%/C% train/ref/val split means
    # an A%/B%/C% split of each class.
    # The split is done on an individual point basis.
    #
    # The POINTS mode's results should already be approximately stratified due
    # to the annotations being shuffled. However, POINTS_STRATIFIED makes the
    # stratification more guaranteed. This can be useful because it makes the
    # final number of unique classes more consistent.
    #
    # Stratification checks that the number of annotations in each
    # set isn't less than the number of unique classes.
    # However, each set is NOT guaranteed to have at least 1 of each class.
    # If stratification is calculated such that a set would get <0.5
    # annotations of a class, then that set gets 0 of that class.
    POINTS_STRATIFIED = 'points_stratified'
```

The mode that's notably 'missing' is `VECTORS_STRATIFIED`, because it would be more complicated to stratify accurately when splitting at the vector level. As @yeelauren pointed out in the old PR's thread, there should be ways to implement that if desired, such as the [imbalanced-learn](https://imbalanced-learn.org/stable/combine.html) library. But it would be more complex to implement than the other modes, so it's deferred until someone really wants it.

There may be other methods/restrictions that one might want for the data split. For example, perhaps you have a hierarchy of CoralNet data where the data can be divided into several sources, each source has a set of feature vectors, and each feature vector has a set of point features; and you want each *source* to go entirely in train, or ref, or val (not split between the three). However, at this point, I think that potential need is covered by the ability to instantiate your own `TrainingTaskLabels` and thus define your own arbitrary split.

Results of experiments using this code:

| Source | Images | Mode              | Annotations | Train   | Ref   | Val    | Classes | Accuracy | CN accuracy | Train time |
| ------ | ------ | ----------------- | ----------- | ------- | ----- | ------ | ------ | -------- | ----------- | ---------- |
| 3342   | 1204   | VECTORS           | 1202766     | 1031819 | 50000 | 120947 | 12     | 95.1%    | 93.0%       | 1267.2s    |
| 3342   | 1204   | POINTS_STRATIFIED | 1203965     | 1033567 | 50000 | 120398 | 16     | 95.7%    | 93.0%       | 1484.2s    |
| 372    | 37955  | VECTORS           | 379383      | 303504  | 37958 | 37921  | 53     | 78.5%    | 78.0%       | 3441.3s    |
| 372    | 37955  | POINTS_STRATIFIED | 379538      | 303628  | 37955 | 37955  | 60     | 78.8%    | 78.0%       | 3323.5s    |
| 2112   | 4649   | VECTORS           | 232263      | 185792  | 23242 | 23229  | 46     | 86.6%    | 82.0%       | 763.9s     |
| 2112   | 4649   | POINTS_STRATIFIED | 232416      | 185930  | 23243 | 23243  | 56     | 86.8%    | 82.0%       | 755.3s     |
| 3401   | 7195   | VECTORS           | 243342      | 194598  | 24399 | 24345  | 60     | 80.0%    | 80.0%       | 1020.7s    |
| 3401   | 7195   | POINTS_STRATIFIED | 243568      | 194851  | 24359 | 24358  | 68     | 80.3%    | 80.0%       | 1123.4s    |
| 3411   | 14948  | VECTORS           | 227448      | 181933  | 22767 | 22748  | 82     | 74.1%    | 74.0%       | 1349.8s    |
| 3411   | 14948  | POINTS_STRATIFIED | 227554      | 182038  | 22758 | 22758  | 88     | 74.7%    | 74.0%       | 1389.1s    |
| 3577   | 3696   | VECTORS           | 184623      | 147635  | 18500 | 18488  | 29     | 89.7%    | 89.0%       | 536.0s     |
| 3577   | 3696   | POINTS_STRATIFIED | 184786      | 147826  | 18480 | 18480  | 36     | 89.3%    | 89.0%       | 554.9s     |
| 1579   | 16438  | VECTORS           | 164334      | 131460  | 16439 | 16435  | 50     | 77.7%    | 77.0%       | 1343.7s    |
| 1579   | 16438  | POINTS_STRATIFIED | 164341      | 131468  | 16437 | 16436  | 48     | 77.3%    | 77.0%       | 1374.4s    |
| 3697   | 1049   | VECTORS           | 52279       | 41810   | 5250  | 5219   | 18     | 78.8%    | 82.0%       | 149.0s     |
| 3697   | 1049   | POINTS_STRATIFIED | 52434       | 41947   | 5244  | 5243   | 25     | 81.5%    | 82.0%       | 159.3s     |
| 3606   | 969    | VECTORS           | 24096       | 19269   | 2425  | 2402   | 42     | 59.4%    | 69.0%       | 138.2s     |
| 3606   | 969    | POINTS_STRATIFIED | 24218       | 19374   | 2422  | 2422   | 49     | 68.4%    | 69.0%       | 127.9s     |
| 3357   | 564    | VECTORS           | 16790       | 13376   | 1710  | 1704   | 14     | 89.4%    | 86.0%       | 74.3s      |
| 3357   | 564    | POINTS_STRATIFIED | 16911       | 13527   | 1692  | 1692   | 15     | 87.3%    | 86.0%       | 77.8s      |
| 3583   | 200    | VECTORS           | 6512        | 5174    | 669   | 669    | 24     | 74.6%    | 77.0%       | 25.1s      |
| 3583   | 200    | POINTS            | 6608        | 5281    | 665   | 662    | 31     | 82.5%    | 77.0%       | 26.3s      |
| 3583   | 200    | POINTS_STRATIFIED | 6637        | 5308    | 666   | 663    | 33     | 84.0%    | 77.0%       | 30.5s      |
| 3362   | 44     | VECTORS           | 1715        | 1327    | 200   | 188    | 5      | 95.2%    | 95.0%       | 7.5s       |
| 3362   | 44     | POINTS_STRATIFIED | 1755        | 1403    | 176   | 176    | 6      | 93.2%    | 95.0%       | 6.9s       |
| 3489   | 86     | VECTORS           | 860         | 680     | 90    | 90     | 10     | 54.4%    | 80.0%       | 7.1s       |
| 3489   | 86     | POINTS_STRATIFIED | 857         | 685     | 86    | 86     | 9      | 59.3%    | 80.0%       | 5.5s       |
| 3685   | 21     | VECTORS           | 580         | 410     | 90    | 80     | 10     | 57.5%    | 46.0%       | 4.0s       |
| 3685   | 21     | POINTS_STRATIFIED | 607         | 484     | 62    | 61     | 13     | 65.6%    | 46.0%       | 3.5s       |

CSV version for potentially easier viewing: 
[2024-03 - single source runs with new sampling code.csv](https://github.com/coralnet/pyspacer/files/14548889/2024-03.-.single.source.runs.with.new.sampling.code.csv)

(To be exact, the experiments used the `training-cache-features-3` branch, which places the PR #80 feature-caching commits on top of this PR's branch).

My takeaways from the experiments:

- The default train/ref/val ratios of 80%/10%/10% (with ref being capped at 50000) are working correctly. Small discrepancies can be explained by 1) restrictions of the VECTORS mode, and 2) filtering out of classes that don't end up in both train and ref.

- Classifiers' measured accuracy hasn't been conclusively affected one way or the other by this PR, compared to the accuracy of pre-existing CoralNet classifiers trained on the same sources. That is, "Accuracy" is comparable to "CN accuracy". There are bigger accuracy fluctuations for smaller sources, as I'd expect.

- POINTS_STRATIFIED consistently results in bigger labelsets ("Classes" column) than VECTORS, as I'd expect, since with POINTS_STRATIFIED, rare classes have a better guarantee of being included in both train and ref.

- And since POINTS_STRATIFIED includes more classes than VECTORS, it also includes slightly more annotations.

- It seems that if there's a discrepancy in accuracy between VECTORS and POINTS_STRATIFIED, larger discrepancies tend to give the nod to POINTS_STRATIFIED, despite the fact that we might expect bigger labelsets to result in lower accuracy (harder to predict correctly when there are more choices). So, for smaller sources in particular, there could indeed be a concern that POINTS_STRATIFIED makes the calibration and validation 'artificially good' by training on the same images that it's validating on. For larger sources, it doesn't seem as much of a concern. (cc @kriegman)

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,12 +204,21 @@ from spacer.messages import TrainingTaskLabels
 from spacer.task_utils import preprocess_labels
 
 # 1
-labels = preprocess_labels(ImageLabels(...), ...)
+labels = preprocess_labels(
+    ImageLabels("see previous code block for example args to ImageLabels..."),
+    "optional args to preprocess_labels()...",
+)
 # 2
 labels = TrainingTaskLabels(
-    train=ImageLabels(...), ref=ImageLabels(...), val=ImageLabels(...))
+    train=ImageLabels(...),
+    ref=ImageLabels(...),
+    val=ImageLabels(...),
+)
 # 3
-labels = preprocess_labels(TrainingTaskLabels(...), ...)
+labels = preprocess_labels(
+    TrainingTaskLabels("args like the previous example..."),
+    "optional args to preprocess_labels()...",
+)
 ```
 
 Once you have a TrainingTaskLabels instance, pass that and the other required arguments to TrainClassifierMsg, and then pass that message to `train_classifier()`, which produces:
--- a/README.md
+++ b/README.md
@@ -190,10 +190,13 @@ image_labels = ImageLabels({
 
 The `labels` argument of `TrainClassifierMsg` expects an instance of `data_classes.TrainingTaskLabels`. There are a few ways to create this:
 
-1. Pass a single ImageLabels instance to the `task_utils.preprocess_labels()` function. preprocess_labels() then decides how to split up your labels into train, ref, and val (while doing error checks in the meantime), and creates a TrainingTaskLabels instance from there.
-2. Pass three ImageLabels instances to the TrainingTaskLabels constructor: one instance for each of train, ref, and val.
-3. Do method 1, but also specify the `accepted_classes` argument to preprocess_labels(); this makes the function filter out any labels that aren't in the desired set of classes.
-4. Do method 2, but also pass the TrainingTaskLabels through preprocess_labels(). This allows you to use the error-checking and accepted_classes parts of preprocess_labels(), and the train/ref/val split you defined will remain intact.
+1. Pass a single ImageLabels instance to the `task_utils.preprocess_labels()` function. preprocess_labels() will:
+   - Split up your labels into train, ref, and val sets; optional arguments are available to control how the split is done.
+   - Do error checks.
+   - Optionally filter out unwanted classes, if you specified the `accepted_classes` argument.
+   - Return a TrainingTaskLabels instance.
+2. Create your own TrainingTaskLabels instance by passing three ImageLabels instances into the constructor: one ImageLabels for each of train, ref, and val. This lets you define your own arbitrary train/ref/val split.
+3. Do method 2, but then pass your TrainingTaskLabels instance through preprocess_labels(). This allows you to use just the error-checking and class-filtering parts of preprocess_labels().
 
 ```python
 from spacer.data_classes import ImageLabels
@@ -201,17 +204,15 @@ from spacer.messages import TrainingTaskLabels
 from spacer.task_utils import preprocess_labels
 
 # 1
-labels = preprocess_labels(ImageLabels(...))
+labels = preprocess_labels(ImageLabels(...), ...)
 # 2
 labels = TrainingTaskLabels(
     train=ImageLabels(...), ref=ImageLabels(...), val=ImageLabels(...))
 # 3
-labels = preprocess_labels(ImageLabels(...), accepted_classes={...})
-# 4
-labels = preprocess_labels(TrainingTaskLabels(...), accepted_classes={...})
+labels = preprocess_labels(TrainingTaskLabels(...), ...)
 ```
 
-So, pass that and the other required arguments to TrainClassifierMsg, and then pass that message to `train_classifier()`, which produces:
+Once you have a TrainingTaskLabels instance, pass that and the other required arguments to TrainClassifierMsg, and then pass that message to `train_classifier()`, which produces:
 
 - A classifier as an `sklearn.calibration.CalibratedClassifierCV` instance, stored in a pickle (.pkl) file. spacer's `storage.load_classifier()` function can help with loading classifiers that were created in older scikit-learn versions. 
 - Classifier evaluation results as a JSON file.
--- a/spacer/task_utils.py
+++ b/spacer/task_utils.py
@@ -1,7 +1,11 @@
 from __future__ import annotations
+from collections import defaultdict
+from enum import Enum
 from logging import getLogger
+import random
 
 from PIL import Image
+from sklearn.model_selection import train_test_split
 
 from spacer import config
 from spacer.data_classes import ImageLabels, LabelId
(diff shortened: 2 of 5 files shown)


StephenChan:
[Stratified Sampling PR](https://trello.com/c/cUUKq0GQ/662-stratified-sampling-pr)

StephenChan:
Oh yeah, one other note: there was a previous version of this code where I made POINTS_STRATIFIED (or equivalent) the default mode. However, the default mode has been changed to VECTORS.

StephenChan:
Yeah, another issue for it - just created issue #98.

Issue #74 also has notes about sklearn's MLP being a bit rudimentary compared to other libraries' implementations. Super robust deep learning implementations seem to be out of sklearn's scope basically.

README.md
```diff
@@ -190,28 +190,29 @@ image_labels = ImageLabels({
...
+3. Do method 2, but then pass your TrainingTaskLabels instance through preprocess_labels(). This allows you to use just the error-checking and class-filtering parts of preprocess_labels().
 
 ```python
 from spacer.data_classes import ImageLabels
 from spacer.messages import TrainingTaskLabels
 from spacer.task_utils import preprocess_labels
 
 # 1
-labels = preprocess_labels(ImageLabels(...))
+labels = preprocess_labels(ImageLabels(...), ...)
```
yeelauren, line 207: Can we be more explicit with these examples ? What goes into ImageLabels(...) and , .... ) ?

spacer/task_utils.py
```diff
@@ -61,102 +65,316 @@ def check_extract_inputs(image: Image,
...
+    # useful.
+    # Thus, this mode can improve usefulness of calibration, and rigor of the
+    # evaluation results.
+    # However, the annotation count may not end up precisely balanced
+    # between train/ref/val as desired, particularly when the feature vector
+    # size is comparable to the set size. For example, if each feature vector
+    # has 100 points, and the target ref-set size is 450, then the best we can
+    # do is giving the ref set either 400 or 500 points.
+    VECTORS = 'vectors'
+    # The split is done on an individual point basis, so a single
```
yeelauren, line 91: Does this mean we are duplicating featurevectors  ?

README.md
```diff
@@ -190,28 +190,29 @@ image_labels = ImageLabels({
 
 The `labels` argument of `TrainClassifierMsg` expects an instance of `data_classes.TrainingTaskLabels`. There are a few ways to create this:
 
-1. Pass a single ImageLabels instance to the `task_utils.preprocess_labels()` function. preprocess_labels() then decides how to split up your labels into train, ref, and val (while doing error checks in the meantime), and creates a TrainingTaskLabels instance from there.
-2. Pass three ImageLabels instances to the TrainingTaskLabels constructor: one instance for each of train, ref, and val.
-3. Do method 1, but also specify the `accepted_classes` argument to preprocess_labels(); this makes the function filter out any labels that aren't in the desired set of classes.
-4. Do method 2, but also pass the TrainingTaskLabels through preprocess_labels(). This allows you to use the error-checking and accepted_classes parts of preprocess_labels(), and the train/ref/val split you defined will remain intact.
+1. Pass a single ImageLabels instance to the `task_utils.preprocess_labels()` function. preprocess_labels() will:
```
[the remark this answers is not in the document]
yeelauren: nevermind see it's provided in the  PR above, maybe a good spot for that table would be in docs we generate

README.md
StephenChan: Sure - the code snippet above this has a full ImageLabels example. Maybe I can point that out in a comment here. The second ... are any optional args for preprocess_labels(); I'll clarify that.

spacer/task_utils.py
```diff
@@ -61,102 +65,316 @@ def check_extract_inputs(image: Image,
                 f" valid range of 0-{im_width - 1}.")
 
 
+class SplitMode(Enum):
+    """
+    How to split annotations between train, ref, and val sets.
+    """
+    # Each feature vector's points all go into train, or all go into ref, or
```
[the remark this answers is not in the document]
StephenChan: Which part(s)? I have to say I was really getting hand-wavy on the 'more different' part, since I don't have a great grasp on the rationale myself. I'll see if I can think of a better way to word that part.

spacer/task_utils.py
StephenChan: No - for example, if you have a feature vector with 20 point-features, then you might have 16 going to train, 2 going to ref, and 2 going to val. I can add an example like that in the comment here.

README.md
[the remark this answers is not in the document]
StephenChan: Yeah, separate issue if anything, but good call - more in-depth docs for decision making would be good in general.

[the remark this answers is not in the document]
StephenChan: Yeah, new issue or separate docs, but good call - more in-depth docs for decision making would be good in general.

README.md
StephenChan: Updated this part.

spacer/task_utils.py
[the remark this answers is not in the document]
StephenChan: Tried making the VECTORS comments more concrete and detailed.

spacer/task_utils.py
StephenChan: Added that example.

yeelauren (approved):

yeelauren:
Great! Thanks @StephenChan. One statistic we're missing here is the _per class_ accuracy. Overall accuracy can hide some of the nuance between classes - opened an issue #99 that should help with this.

PR #104, opened 2024-08-16
Allow using AWS metadata service when available.

michaelconnor00:
This Draft PR is to provide an example of using credentials when running in AWS where a role is used and credentials are provided through the metadata service. This is how the credentials are managed for all resources in the Mermaid API. 

This is yet to be tested, but I wanted to document how it can be done. I will test it on Monday.


saanobhaai:
The `resource()` call requires a service name, i.e. it should be
`THREAD_LOCAL.s3_connection = boto3.resource("s3")`

StephenChan:
Updated version: PR #105

PR #105, opened 2024-09-06
AWS credentials through metadata service or boto auto-detection

StephenChan:
- Metadata service (STS): This is like PR #104 from @michaelconnor00 with the addition of catching `botocore.exceptions.NoCredentialsError`, which gets raised when credentials can't be obtained this way.

- Boto auto-detection: After having another overall look at `get_s3_conn()`, I noticed that the comment of

  > If [spacer-config] credentials are None, it will default to using credentials in ~/.aws/credentials

  didn't actually hold, because the spacer code would raise a `ConfigError` if spacer-config credentials were None. I decided to just remove that check that would lead to the ConfigError. The check additionally seemed a bit misleading anyway, because it's no guarantee of having valid credentials; the `boto3.resource()` call doesn't actually use the credentials right away. You'll only know if they're valid credentials when the resource (or 'connection', as we currently call it) is actually used to access an S3 bucket or object for the first time.

Will take another look at this later to make sure I didn't miss something else.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## 0.9.1 (WIP)
+
+- AWS credentials can now be obtained through the following methods, in addition to spacer config values as before:
+  
+  - AWS's metadata service (STS); proof of concept from @michaelconnor00
+  - boto's auto-detection logic when neither of STS or spacer config are used (this was intended to work before, but needed fixing)
+
 ## 0.9.0
 
 - Python 3.8 and 3.9 support have been dropped; Python 3.11 support has been added.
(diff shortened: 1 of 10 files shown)


michaelconnor00:
Looks good to me. 

Closes #104

StephenChan:
Thanks. I just renamed `get_s3_conn()` to `get_s3_resource()` and added unit tests, which should be enough for this PR.

I'll address the CI failing check in another PR; looks to be related to numpy 2.x being allowed by the package requirements.

After that I'll release a new pyspacer version, and let me know if any other issues arise. For example, I don't know what it looks like when an STS-retrieved resource expires (I assume that is something that happens eventually).

PR #111, opened 2024-12-03
Feature extractor refactoring + more granular test-related config

StephenChan:
My main goal going into this PR was to make the extractor-related unit tests less dependent on AWS. That way, GitHub CI could run more of those tests, so that the next round of package updates could be more thoroughly tested with GitHub CI (with multiple combos of package versions, which is onerous to test outside of CI).

To do that, I wanted to come up with some code that gave valid EfficientNet extractor weights, without actually having to load CoralNet's weights from its AWS infrastructure. The result of that is the `EfficientNetExtractor.untrained_instance()` method in this PR.

To get there, I found it easier to refactor out some of the `pyparams` passing which was going on in the extractor classes/functions, and this entailed moving most of the non-class functions into the Extractor classes. That naturally led to related refactors. I also ended up aiming farther than the original goal regarding unit test dependencies.

Overall, I'll summarize with what I added to the CHANGELOG:

- Feature extractor class changes:

  - `FeatureExtractor` and its built-in subclasses should now be imported like `from spacer.extractors import <class>` instead of `from spacer.extract_features import <class>`.

  - High-level usage of `FeatureExtractor` instances is the same as before - invoking `__call__()` performs feature extraction on an image. However, subclass implementations should now generally define a `patches_to_features()` method instead of overriding `__call__()`.

  - There is now a `TorchExtractor` class which has details that are specific to PyTorch but not to EfficientNet. So, it's suitable as a starting point for a custom PyTorch extractor that uses another type of network. `EfficientNetExtractor` now inherits from TorchExtractor.

  - There are now `CROP_SIZE` and `BATCH_SIZE` class-level variables available.

- Config and test changes:

  - Some former usages of `TEST_BUCKET` have been changed to `CN_FIXTURES_BUCKET`, to more clearly denote test fixtures that are currently only available to CoralNet devs.

  - The remaining usages of `TEST_BUCKET` are now usable by anyone with an AWS account. This can be any S3 bucket that you have read and write access to.

  - `TEST_EXTRACTORS_BUCKET` is now known as `CN_TEST_EXTRACTORS_BUCKET`, again denoting fixtures currently only available to CoralNet devs.

  Related to these changes, now more tests are runnable without needing CoralNet AWS credentials. More tests are runnable in GitHub Actions CI, as well (even though that doesn't use AWS at all).


StephenChan:
And the result in CI is: 42 skipped tests out of 193, whereas it was previously 50 skipped tests out of 194 (PR #110).

PR #113, opened 2024-12-04
scikit-learn 1.5.2, numpy 2, Python 3.12

StephenChan:
It's time for dependency updates; Dependabot alerts are piling up a bit.

I thought scikit-learn would have been the main one to watch out for in terms of compatibility, but it was actually the combo of numpy and torch which was trickiest. torch 2.4 made a move towards better safety of `torch.load()` by issuing a warning if general-purpose unpickling was used, as opposed to a limited unpickling mode which is less powerful (thus safer) but may need adding to an allow-list to load everything we need. So that's what I had to do in commit d8b5f41, with the additional complication that one of the things to be loaded was from `numpy.core`, which got renamed to `numpy._core` in numpy 2.

The CHANGELOG has a couple of notes about compatible versions:

> If you use numpy>=2.0, you probably also need torch>=2.3, torchvision>=0.18, and scipy>=1.13 (scipy is required by scikit-learn). Otherwise, you may get errors like "_ARRAY_API not found" and warnings like "A module that was compiled using NumPy 1.x cannot be run in NumPy 2.1.3 as it may crash."
>
> Also note that torch>=2.3 does not provide binaries for macOS x86.

There are still no issues with building Caffe, even on numpy 2.

By updating everything, it looks like we're able to support Python 3.12 in the meantime.

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -19,7 +19,7 @@ jobs:
             requirements: "dev"
           - python-version: "3.10"
             requirements: "min"
-          - python-version: "3.11"
+          - python-version: "3.12"
             requirements: "max"
 
     steps:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,16 +12,6 @@
 
   - There are now `CROP_SIZE` and `BATCH_SIZE` class-level variables available.
 
-- Config and test changes:
-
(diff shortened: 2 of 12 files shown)


StephenChan:
I did some additional testing to confirm that I haven't accidentally changed the behavior of existing extractors. I ran the CoralNet vision-backend jobs on 4 test sources, 2 EfficientNet and 2 VGG16 (ranging from 60 to 10,000 confirmed images each). The resulting classifiers for each of the 4 sources all ended up with the same accuracy percentages as the test runs I did on the same data in May 2024.

PR #118, opened 2025-04-19
PyTorch 2.6

StephenChan:
This puts torch and torchvision at the latest versions, 2.6 and 0.21, to address a security alert regarding older torch versions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Changelog
 
-## 0.11.1 (WIP)
+## 0.12.0 (WIP)
+
+- Updates to pip-install dependencies:
+
+  - torch: >=2.2,<2.5 to >=2.6,<2.7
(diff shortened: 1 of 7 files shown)


StephenChan:
The `AttributeError: module 'numpy' has no attribute 'dtypes'` addressed in the second commit only revealed itself now because it needed numpy<1.25 and torch>=2.4 to happen, and that combo of versions is now in `min.txt`. The solution was to add

```python
if hasattr(np, 'dtypes'):
    safe_globals.append(np.dtypes.Int64DType)
```

There is no else clause, since in this combo of versions we apparently don't need to add an alternative safe-global.

StephenChan:
I am now noticing another security alert for all torch releases (i.e. <=2.6.0) with moderate severity, but I guess nothing reasonable can be done about that right now, and at least this is a downgrade from the previous critical severity...

PR #137, opened 2026-02-27
Allow generalized ClassifierTrainer inheritance, keeping backwards compatibility with the MiniBatchTrainer

gnieuwenhuis:
## Goal
To allow more control over the training loop in the mermaid classifier (see for example https://github.com/data-mermaid/mermaid-classifier/pull/18)

To accomplish this, we are allowing inheritance of the ClassifierTrainer here. This in turn allows defining custom training logic within an inherited trainer class. 

See this discussion for more background: https://github.com/data-mermaid/mermaid-classifier/issues/17

## These changes include
- make it possible to inherit from the ClassifierTrainer, and use that inherited trainer instead of the default MiniBatchTrainer
- If you do not specify a custom trainer, behaviour remains unchanged, and the default MiniBatchTrainer is still used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,7 +261,7 @@ message = TrainClassifierMsg(
     # For your bookkeeping.
     job_token='classifier1',
     # 'minibatch' is currently the only trainer that spacer defines.
-    trainer_name='minibatch',
+    trainer='minibatch',
     # How many iterations the training algorithm should run; more epochs
     # = more opportunity to converge to a better fit, but slower.
     nbr_epochs=10,
--- a/README.md
+++ b/README.md
@@ -260,8 +260,10 @@ from spacer.task_utils import preprocess_labels
 message = TrainClassifierMsg(
     # For your bookkeeping.
     job_token='classifier1',
-    # 'minibatch' is the default trainer defined by spacer.
-    # Custom trainers can be created by subclassing ClassifierTrainer.
+    # 'minibatch' specifies spacer's default trainer, MiniBatchTrainer,
+    # with default parameters.
+    # For more control, you can instead pass a MiniBatchTrainer instance, or
+    # define your own ClassifierTrainer subclass and pass an instance of it.
     trainer='minibatch',
     # How many iterations the training algorithm should run; more epochs
     # = more opportunity to converge to a better fit, but slower.
--- a/README.md
+++ b/README.md
@@ -260,7 +260,8 @@ from spacer.task_utils import preprocess_labels
 message = TrainClassifierMsg(
     # For your bookkeeping.
     job_token='classifier1',
-    # 'minibatch' is currently the only trainer that spacer defines.
+    # 'minibatch' is the default trainer defined by spacer.
+    # Custom trainers can be created by subclassing ClassifierTrainer.
     trainer='minibatch',
     # How many iterations the training algorithm should run; more epochs
     # = more opportunity to converge to a better fit, but slower.
--- a/spacer/config.py
+++ b/spacer/config.py
@@ -212,6 +212,8 @@ CLASSIFIER_TYPES = [
     'MLP'
 ]
 
+# Deprecated: use ClassifierTrainer instances or trainer_factory() instead.
+# Kept for backward compatibility with code that checks membership.
 TRAINER_NAMES = [
     'minibatch'
 ]
--- a/spacer/messages.py
+++ b/spacer/messages.py
@@ -189,8 +189,10 @@ class TrainClassifierMsg(DataClass):
         self,
         # For caller's reference.
         job_token: str,
-        # 'minibatch' is currently the only trainer that spacer defines.
-        trainer_name: str,
+        # A ClassifierTrainer instance, or a string that resolves to one
+        # via trainer_factory() (e.g. 'minibatch' or a fully-qualified
+        # class path like 'mypackage.trainers.MyTrainer').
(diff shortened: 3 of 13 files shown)


StephenChan:
Thanks @gnieuwenhuis - looks great all around.

- I like how you added not just the ClassifierTrainer instance option for `trainer`, but also the class path option, taking advantage of the fact that ClassifierTrainer doesn't take any init parameters (unlike FeatureExtractor). This enables a user to specify a `trainer` without having to import the trainer class first.

- Nice to see how support for the old `trainer_name` message parameter is retained as well. For the record, I actually have been less careful with changing the pyspacer message APIs, justifying a lack of deprecation cycles with pyspacer's "0.x" version numbering (something 1.0 or above would indicate a more stable package). But I think it's good to get into this habit of compatibility between consecutive pyspacer versions... the only reason it's not a problem for CoralNet is that we just accept some downtime during server updates, which lets us finish all submitted pyspacer jobs before updating the code. Anyway, this is good to have!

- The ClassifierTrainer changes themselves look good, and consistent with FeatureExtractor. Good idea on adding this check. I haven't actually thought of using `isinstance(<something>, type)` before.

  ```python
  if not (
      isinstance(trainer_class, type)
      and issubclass(trainer_class, ClassifierTrainer)
  )
  ```

- Defining the built-in trainers in `train_classifier.py` instead of `config.py`: also good IMO.

- The tests look complete and clean. Probably worth copying many of these test ideas into the feature extractor tests later.

I will run some tests myself, including with CoralNet, to check that everything's good to go. Otherwise the only to-do point on my mind is on `README.md`; the comment saying "'minibatch' is currently the only trainer that spacer defines." should be expanded to briefly explain that custom trainers are possible too.

StephenChan:
Thanks for the README update. I also see a `CLAUDE.md` that you might've added accidentally though?

Testing on my end was all good.

One other comment I forgot to make: I was originally thinking of moving one or more of train(), evaluate_classifier(), and calc_acc() from train_utils.py into ClassifierTrainer or MiniBatchTrainer. However, at this point it's not entirely clear how that would increase ease of use, particularly when there is still only one built-in trainer class. So for now, I think it makes sense to not fix what's not broken.

gnieuwenhuis:
Sorry about the added file - it was mistakenly committed and I'll remove right away. I've been using AI to help explore, ideate, and suggest solutions lately, which is why it was there to begin with.

I see your idea with the train(), evaluate_classifier() and calc_acc(), and I thought about it a little just now. For what its worth, I think that the evaluate_classifier() and calc_acc() are both functions that act more like building blocks which are easily useable in different ways if they remain in train_utils.py. The train() method is core functionality to the MiniBatchTrainer, and I think it would improve class cohesion to put it with that trainer. But I wouldn't change the abstract ClassifierTrainer to avoid overly constraining custom trainers (my thought is a little selfish, because I'm thinking about how I'm defining the mermaid-classifier custom trainer, and prefer the freedom that is available right now).

I'm happy making some modifications here if you want to to do that - my guess is that if it doesn't get done now it may not get done in the future either.

StephenChan:
> Sorry about the added file - it was mistakenly committed and I'll remove right away. I've been using AI to help explore, ideate, and suggest solutions lately, which is why it was there to begin with.

No worries!

> ... The train() method is core functionality to the MiniBatchTrainer, and I think it would improve class cohesion to put it with that trainer. ... I'm happy making some modifications here if you want to to do that - my guess is that if it doesn't get done now it may not get done in the future either.

Good points here, and thanks - I would like to see this if you're up for doing it.

gnieuwenhuis:
I've moved that function over, and also allowed setting some of the hyperparameters that were previously hard coded - to me it seemed like a good addition for anyone experimenting with the model in the future, but if you prefer not to expose those let me know.

StephenChan:
Great idea, the parameters are certainly nice to have. Thanks!

Most of the tests look good, but `test_custom_batch_size()` might be missing the mark - the length of `ref_acc` verifies the number of epochs, but not the number of batches. Those two distinct numbers just coincidentally match in this test.

To assert on the number of batches, one idea is to use the `spy_decorator()` helper which is currently in `test_tasks.py`. The helper's probably best relocated to `tests/utils.py` instead of that file. Then I think something like this could be done:

```python
load_spied = spy_decorator(ImageLabels.load_data_in_batches)
with mock.patch.object(ImageLabels, 'load_data_in_batches', load_spied):
    ...    # Call trainer._train()
load_spied.mock_obj.assert_called_with(batch_size=50, random_seed=1)
```

Alternatively one could spy on `partial_fit()` to check its `call_count`. Either way is good IMO.

gnieuwenhuis:
Nice catch! Check is fixed using the spy_decorator as you suggested, seems like a handy helper.

StephenChan:
Great! Thanks. Lastly, I have a couple of comment updates/additions in mind. I might as well write that myself and commit it.

StephenChan:
@gnieuwenhuis Anything else in mind before merging this?

After merging this, I can also release a new pyspacer version if that'll be useful soon.

gnieuwenhuis:
Nothin more from me - I'm happy if you are.

It would be useful if you can release a new version, I can pin it in my draft PR's on the classifier and start moving forward with those

StephenChan:
Sounds good, I'll let you know when the release is made.

StephenChan:
@gnieuwenhuis 0.14.0 release has been made.

